PR #6, opened 2021-09-25
Add argument for setting conan options

aminya:


aminya (commented):
This should be fixed. The last time I tried it was not working.

PR #8, opened 2021-09-26
Feature/use cpm cmake module

ClausKlein:
draft only to open a discussion


aminya (commented):
This needs a rebase

src/Index.cmake
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.16)
 
-set(CMAKELIB_SRC_DIR ${CMAKE_CURRENT_LIST_DIR})
+set(CMAKELIB_SRC_DIR ${CMAKE_CURRENT_LIST_DIR} CACHE FILEPATH "")
```
aminya, line 3: This PR title doesn't match the change. Am I missing something here?

aminya (changes requested):
I had many issues with using CMakeLists for a CMake library. It broke Conan dependencies the last time I tried it.

ClausKlein:
> I had many issues with using `CMakeLists.txt` for a CMake library. It broke Conan dependencies the last time I tried it.

I will have a look at this

PR #37, opened 2021-12-13
fix: rename ProjectOptions to project_options

aminya:
BREAKING CHANGE now, you should rename any usage of ProjectOptions to project_options

This was done for consistency with CMake's naming convention


aminya:
@lefticus I have lost my access to the repository settings, and I cannot rename the repository.

lefticus:
You should have full admin access now @aminya

PR #41, opened 2022-01-20
fix: add EXISTS keyword to BuildType check

### LINKED ISSUE #40: Default build type not set
Hi,

i'm not shure if i got an issue here or if it is just me beeing stupid. 

It seems that the default build type is not correctly set. When i configure a project created from the cpp-starter-project template and just configure it, i do not get the message  `"Setting build type to 'RelWithDebInfo' as none was specified."` from `StandardProjectSettings.cmake:3`.

I injected a message to check if those variables are already set to something:
```
message(STATUS "Got CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}' or CMAKE_CONFIGURATION_TYPE: '${CMAKE_CONFIGURATION_TYPE}'")
``` 
This got me the following message:
```
[cmake] -- Got CMAKE_BUILD_TYPE: '' or CMAKE_CONFIGURATION_TYPE: ''
```

Both variables seem to be empty. But why is the if condition in `StandardProjectSettings.cmake:2` not evaluated to true?

I'm on Windows 10 using the Visual Studio 2019  BuildTools with the MSVC Version 19.29.30037.0.

This furthermore leads to conan fail because the BuildType is not consistent.

LtdSauce:
This commit adds the `EXISTS` keyword in the if-statement so that a missing
BUILD_TYPE is realy detected.

This should close #40

### DIFF
--- a/src/StandardProjectSettings.cmake
+++ b/src/StandardProjectSettings.cmake
@@ -1,5 +1,5 @@
 # Set a default build type if none was specified
-if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+if(NOT EXISTS CMAKE_BUILD_TYPE AND NOT EXISTS CMAKE_CONFIGURATION_TYPES)
   message(STATUS "Setting build type to 'RelWithDebInfo' as none was specified.")
   set(CMAKE_BUILD_TYPE
       RelWithDebInfo


aminya:
Is not `EXISTS` a keyword for files?

aminya (changes requested):
Is not `EXISTS` a keyword for checking the existence of files on the system? 
https://cmake.org/cmake/help/latest/command/if.html#exists

LtdSauce:
Huh... you are right... i rushed to conclusions there. 

With my change it always assumes the default because those files do not exist. Sorry!

aminya:
> Huh... you are right... i rushed to conclusions there.
> 
> With my change it always assumes the default because those files do not exist. Sorry!

Yes. If you want to be more precise, you can change these to check for the empty strings using `NOT "${var}" STREQL ""`

LtdSauce:
You suggestion resulted in a weird error:
```
CMake Error at build/_deps/_project_options-src/src/StandardProjectSettings.cmake:2 (if):
  if given arguments:

    "NOT" "CMAKE_BUILD_TYPE" "STREQL" "" "AND" "NOT" "CMAKE_CONFIGURATION_TYPES" "STREQL" ""

  Unknown arguments specified
  ```
  
  Actually i had a litte typo in my diagnostic message. Changing `CMAKE_CONFIGURATION_TYPE` to `CMAKE_CONFIGURATION_TYPES` results in output. This variable is not empty:
  ```
  Got CMAKE_BUILD_TYPE: '' or CMAKE_CONFIGURATION_TYPE: 'Debug;Release;MinSizeRel;RelWithDebInfo'
  ```
  
  Yet i do not understand, why this variable disables the default? With `CMAKE_CONFIGURATION_TYPES` beeing not empty but `CMAKE_BUILD_TYPE` beeing empty my configure fails because conan wants `CMAKE_BUILD_TYPE`to be set:
  ```
  [cmake] -- Conan: Adding bincrafters remote repository (https://bincrafters.jfrog.io/artifactory/api/conan/public-conan) verify ssl (True)
[cmake] -- Multi-configuration build: 'Debug;Release;MinSizeRel;RelWithDebInfo'!
[cmake] -- Running Conan for build type 'Debug'
[cmake] CMake Error at build/conan.cmake:71 (message):
[cmake]   Please specify in command line CMAKE_BUILD_TYPE
[cmake]   (-DCMAKE_BUILD_TYPE=Release)
[cmake] Call Stack (most recent call first):
[cmake]   build/conan.cmake:430 (_conan_detect_build_type)
[cmake]   build/_deps/_project_options-src/src/Conan.cmake:39 (conan_cmake_autodetect)
[cmake]   build/_deps/_project_options-src/src/Index.cmake:174 (run_conan)
[cmake]   CMakeLists.txt:22 (project_options)
```

Is this related?

LtdSauce:
I'd propose only checking `CMAKE_BUILD_TYPE` to determine if the default should be used. Or maybe it is better to use `OR` instead of `AND`?

LtdSauce:
As of [CMAKE_CONFIGURATION_TYPES Documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html) a call to `project()` will initialize this variable to the default of the generator. Same is done to `CMAKE_BUILD_TYPE`. 
It seems that the Visual Studio 2019 generator sets the value of `CMAKE_CONFIGURATION_TYPES` to "something" while setting `CMAKE_BUILD_TYPE` to the empty string. 

Because the call to `project_options` is after the call to `project` this will always fail.

aminya (approved):
This looks good. Yeah, I do not think checking the configuration type is needed. We can get the input from @lefticus too.

LtdSauce:
I'm getting some doubts that my change maybe disables the multibuild capabilities of Visual Studio and Multi Ninja. I try to play around with it as soon as i can.

Probably thats why the BUILD_TYPE is not Set to anything, because Visual Studio uses Multi build configuration?

LtdSauce:
My findings in #40 suggest, that this PR should be closed. It looks like it is an issue with the conanfile.txt from the cpp_starter_project

LtdSauce:
I close my PR. It does not fix the problem that it should. On multi config builds it still has no effect.

aminya:
Thanks for the exploration!

PR #43, opened 2022-01-23
Fixing set_project_warnings() function

avimalka:
WARNINGS_AS_ERRORS is not working.
WARNINGS_AS_ERRORS gets the value "WARNINGS_AS_ERRORS=TRUE" instead of just "TRUE" (at src/CompilerWarnings.cmake:72) and thus does not add -Werror
Same applies to MSVC/GCC/CLANG_WARNINGS


aminya (approved):
At some point we where using CMake's arge parse. That's why we forgot to change it.

PR #46, opened 2022-01-27
Add dynamic_project_options + developer mode + other fixes

lefticus:
* add utility function for getting all targets (necessary for ASAN improvements with Windows in cpp_starter_project)
 * add new `dynamic_project_options` utility that provides the user with a set of recommended defaults
    * defaults can be overridden with variables before calling `dynamic_project_options`
    * defaults to recommended levels of analysis per platform
    * adds a new "Developer Mode" and "User Mode"
       * Developer Mode: has all recommended analysis settings, is the default
       * User Mode: disables runtime analysis, static analysis, and warnings_as_errors, enabled with a single switch for people who just want to compile and run the program with no extra tools getting in their way
    * all options are configurable with CMake UIs
 * refine cppcheck options so common issues like "missing includes" do not cause failures
 * configure-time error if the user tries to enable PCH on GCC with clang-tidy, since clang-tidy cannot parse GCC's PCH files


src/RecommendedOptions.cmake
```diff
@@ -0,0 +1,130 @@
...
+# applied options from the GUI which are set
+
+# Any default can be overridden
+# set(<feature_name>_DEFAULT <value>) - set default for both user and developer modes
+# set(<feature_name>_DEVELOPER_DEFAULT <value>) - set default for developer mode
+# set(<feature_name>_USER_DEFAULT <value>) - set default for user mode
+
+option(ENABLE_DEVELOPER_MODE "Set up defaults for a developer of the project, and let developer change options" ON)
+
+if(MSVC)
```
bhardwajs, line 24: How would this behave when using `clang` on Windows? Should this be `if(WIN32)`?

lefticus: I believe these checks are correct. If you're not using MSVC, you're using a different generator, (Make or Ninja likely) and these things should all work as expected.

The bigger question I had, is, should this include all of the non-Make/Ninja generators, since those do not support clang-tidy et al.

bhardwajs: https://cmake.org/cmake/help/latest/variable/MSVC.html?highlight=msvc
Suggests that's not the case. Especially, if one is using Ninja generator on Windows with environment variables `CC=cl.exe` and `CXX=cl.exe`

I am with you that this should not include generators that don't support `clang-tidy`.

lefticus: OK thanks, I'll update this to be more explicit based on generator, not compiler.

src/RecommendedOptions.cmake
```diff
@@ -0,0 +1,136 @@
...
+    ${option_user_default})
+
+  if(OPT_${option_name})
+    set(${option_name}_VALUE ${option_name})
+  else()
+    unset(${option_name}_VALUE)
+  endif()
+endforeach()
+
+macro(recommended_project_options)
```
aminya, line 115: Should we change this to `dynamic_project_options`? This implies that `project_options` is not recommended or has bad defaults! 

```suggestion
macro(dynamic_project_options)
```

src/StaticAnalyzers.cmake
```diff
@@ -3,8 +3,7 @@ macro(enable_cppcheck)
   if(CPPCHECK)
     set(CMAKE_CXX_CPPCHECK
         ${CPPCHECK}
-        --suppress=missingInclude
-        --enable=all
+        --enable=style
```
aminya, line 6: I don't think we should limit the checks only to style. The default settings looked fine to me. Unless there is a bug that I haven't faced. It actually finds some errors that no other tool can find.

src/RecommendedOptions.cmake
```diff
@@ -0,0 +1,136 @@
...
+
+# Any default can be overridden
+# set(<feature_name>_DEFAULT <value>) - set default for both user and developer modes
+# set(<feature_name>_DEVELOPER_DEFAULT <value>) - set default for developer mode
+# set(<feature_name>_USER_DEFAULT <value>) - set default for user mode
+
+option(ENABLE_DEVELOPER_MODE "Set up defaults for a developer of the project, and let developer change options" ON)
+
+if((CMAKE_CXX_COMPILER_ID MATCHES ".*Clang.*" OR CMAKE_CXX_COMPILER_ID MATCHES ".*GNU.*") AND NOT WIN32)
+  set(GCC_OR_CLANG ON)
```
aminya, line 25:
```suggestion
  set(SUPPORTS_UBSAN ON)
```

src/RecommendedOptions.cmake
```diff
@@ -0,0 +1,136 @@
...
+  set(MAKEFILE_OR_NINJA ON)
+else()
+  set(MAKEFILE_OR_NINJA OFF)
+endif()
+
+include(CMakeDependentOption)
+
+# <option name>;<user mode default>;<developer mode default>;<description>
+set(options
+    "ENABLE_CACHE\;${MAKEFILE_OR_NINJA}\;${MAKEFILE_OR_NINJA}\;Enable ccache on Unix"
```
aminya, line 40: Why should we disable cache on Windows?

src/RecommendedOptions.cmake
```diff
@@ -0,0 +1,136 @@
...
+    "WARNINGS_AS_ERRORS\;OFF\;ON\;Treat warnings as Errors"
+    "ENABLE_CLANG_TIDY\;OFF\;${MAKEFILE_OR_NINJA}\;Enable clang-tidy analysis during compilation"
+    "ENABLE_CONAN\;ON\;ON\;Automatically integrate Conan for package management"
+    "ENABLE_COVERAGE\;OFF\;OFF\;Analyze and report on coverage"
+    "ENABLE_SANITIZER_ADDRESS\;OFF\;ON\;Make memory errors into hard runtime errors (windows/linux/macos)"
+    "ENABLE_SANITIZER_UNDEFINED_BEHAVIOR\;OFF\;${GCC_OR_CLANG}\;Make certain types (numeric mostly) of undefined behavior into runtime errors"
+    "ENABLE_CPPCHECK\;OFF\;${MAKEFILE_OR_NINJA}\;Enable cppcheck analysis during compilation"
```
aminya, line 47: Why does not cppcheck work with Visual Studio?

aminya (changes requested):

src/GlobalOptions.cmake
```diff
@@ -1,27 +0,0 @@
-set(args
```
aminya, line 1: Now that this one is removed, we should update the last readme example accordignly.

aminya (changes requested):

src/RecommendedOptions.cmake
lefticus: It sets the recommended defaults, which is to have conan enabled. And it actually does make sense to allow the developer to disable conan, if they want to run it manually, which, IIRC is the entire reason we moved to a conanfile.txt.

How about `recommended_default_project_options`

src/StaticAnalyzers.cmake
lefticus: Those settings didn't actually work. You cannot suppress missingInclude for all files on the command line, so that generated its own warnings.

According to the cppcheck manpage, `--style` enables all of the checks that we want.

image

src/RecommendedOptions.cmake
lefticus: This doesn't disable it on windows, it requires Makefiles and Ninja, according to CMake's documentation:

image

src/RecommendedOptions.cmake
lefticus: Again, it's a question of Makefiles or Ninja generators.

Unless there's something incorrect with CMake's documentation here

image

src/StaticAnalyzers.cmake
lefticus: I've explicitly added the other options, just in case the manpage is wrong.

src/RecommendedOptions.cmake
lefticus: I have made the variable names and checks much more explicit

src/RecommendedOptions.cmake
lefticus: Changed.

src/RecommendedOptions.cmake
lefticus: Updated code documentation to reflect why this default was chosen.

src/RecommendedOptions.cmake
lefticus: Updated code documentation to reflect why this default was chosen.

src/RecommendedOptions.cmake
aminya: > It sets the recommended defaults

And also adds dynamic global options for all the function parameters that can be changed from CLI or GUI. That's the part I have problems with. We could instead use local variables, which does not affect the reproducibility.

> And it actually does make sense to allow the developer to disable conan, if they want to run it manually, which, IIRC is the entire reason we moved to a conanfile.txt.

As far as I understand, a project either uses Conan or does not. If it does, then the option should be written in the CMake file. The reason for using conanfile.txt is that it is "declarative" like all the package managers (e.g. cargo, npm, etc).

lefticus: > As far as I understand, a project either uses Conan or does not. If it does, then the option should be written in the CMake file. The reason for using conanfile.txt is that it is "declarative" like all the package managers (e.g. cargo, npm, etc).

Incorrect. We are using the cmake package finders. Those packages can come from

 * conan run automatically by cmake
 * conan run manually by the user
 * system provided packages

lefticus: > And also adds dynamic global options for all the function parameters that can be changed from CLI or GUI. That's the part I have problems with. We could instead use local variables, which does not affect the reproducibility.

I do not see how this affects reproducibility. It gives developers flexibility to change the options they need during the development process. None of these options fundamentally change the nature of your C++ project.

They also allow a user of the program to trivially easily test building on an unsupported compiler by setting the high level option of `DEVELOPER_MODE` to `Off`. This helps in the cases where an untested compiler introduces different warnings, and does not force the user to manually edit the CMakeLists or fix the warnings before they can use the application.

aminya: Could you update the readme example, so we can see how this comes together?

src/CompilerWarnings.cmake
```diff
@@ -69,8 +69,7 @@ function(
     )
   endif()
 
-  if(WARNINGS_AS_ERRORS STREQUAL TRUE)
-    message(AUTHOR_WARNING "NOTE: WARNINGS_AS_ERRORS=${WARNINGS_AS_ERRORS}")
+  if(WARNINGS_AS_ERRORS STREQUAL "TRUE")
```
ddalcino, line 72: Is “TRUE” the only acceptable positive value here? What if the user sets WARNINGS_AS_ERRORS to “ON” or “1”?

According to https://cmake.org/cmake/help/latest/command/if.html#basic-expressions, you can use `if(WARNINGS_AS_ERRORS)` to check for any of the ‘true’ constants, which are “ 1, ON, YES, TRUE, Y, or a non-zero number”. I’m not sure if you need the `${}`.

lefticus: Good point, it got caught up in the fixes I was making to fix the warning overrides.

lefticus: Good point, it got caught up in the fixes I was making to fix the warning overrides.

lefticus: 584f03f

src/GlobalOptions.cmake
lefticus: resolved in a7eb218

lefticus:
@aminya I cannot see any way to actually mark your review as resolved, since it's on a file that was deleted. I can "dismiss" it, but opted to just request you re-review

README.md
```diff
@@ -157,43 +171,16 @@ include(${_project_options_SOURCE_DIR}/src/GlobalOptions.cmake)
...
-      # ENABLE_UNITY
-      # ❗ Now, the address and undefined behavior sanitizers are enabled through CMake options
-      ${ENABLE_SANITIZER_ADDRESS}
-      ${ENABLE_SANITIZER_UNDEFINED_BEHAVIOR}
-      # ENABLE_SANITIZER_LEAK
-      # ENABLE_SANITIZER_THREAD
-      # ENABLE_SANITIZER_MEMORY
+dynamic_project_options(
+  # set PCH headers you want enabled. Format can be slow, so this might be helpful
+  PCH_HEADERS <vector> <string> <fmt/format.h>
```
aminya, line 183: Personally, I am not a fan of PCH headers especially now that modules support is being improved.  PCH headers usually fail on me. Is it really recommended to use them by default? 

```suggestion
  # PCH_HEADERS <vector> <string> <fmt/format.h>
```

src/DynamicProjectOptions.cmake
```diff
@@ -0,0 +1,146 @@
...
+    ${ENABLE_DOXYGEN_VALUE}
+    ${ENABLE_USER_LINKER_VALUE}
+    ${ENABLE_BUILD_WITH_TIME_TRACE_VALUE}
+    ${ENABLE_UNITY_VALUE}
+    ${ENABLE_SANITIZER_ADDRESS_VALUE}
+    ${ENABLE_SANITIZER_LEAK_VALUE}
+    ${ENABLE_SANITIZER_UNDEFINED_BEHAVIOR_VALUE}
+    ${ENABLE_SANITIZER_THREAD_VALUE}
+    ${ENABLE_SANITIZER_MEMORY_VALUE}
+    ${ARGN})
```
aminya, line 145: Does this ARGN allow writing the options in the CMake file like how it is done for `project_options`? If that's the case, it is worth mentioning in the readme example.

README.md
```diff
@@ -148,7 +162,7 @@ FetchContent_MakeAvailable(_project_options)
 include(${_project_options_SOURCE_DIR}/Index.cmake)
 
  # ❗ Add global CMake options
```
aminya, line 164:
```suggestion
 # ❗ Add dynamic CMake options
```

README.md
lefticus: Above it just says "this is an example of what is possible" it is not the default. https://github.com/cpp-best-practices/project_options/pull/46/files/a7eb218f05a2ee9ef6a9d4074a67f1e76cb9f90f#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R175

But a few notes:

 * CMake's PCH support is extremely stable. I have not had a problem with it in the past couple of years
 * In the example I only set it as the default for User Build Mode, not Developer mode. The problem with it being set in developer mode is that it globally adds a set of header files and developers start to accidentally rely on that
 * The CI does test Developer and User build modes, so any accidental breakage will be caught

src/DynamicProjectOptions.cmake
lefticus: It's there for the `PCH_HEADERS` options and `*_WARNING` overrides. If other overrides of options work, it is by accident.

lefticus:
I think this is ready to go and all changes have been addressed (indirectly in some cases). Notably: PCH is never enabled by default. That part of the README is documented to just be *one thing* that is possible to override the defaults on.

src/DynamicProjectOptions.cmake
```diff
@@ -0,0 +1,152 @@
...
+    "WARNINGS_AS_ERRORS\;OFF\;ON\;Treat warnings as Errors"
+    "ENABLE_CLANG_TIDY\;OFF\;${MAKEFILE_OR_NINJA}\;Enable clang-tidy analysis during compilation"
+    "ENABLE_CONAN\;ON\;ON\;Automatically integrate Conan for package management"
+    "ENABLE_COVERAGE\;OFF\;OFF\;Analyze and report on coverage"
+    "ENABLE_SANITIZER_ADDRESS\;OFF\;${SUPPORTS_ASAN}\;Make memory errors into hard runtime errors (windows/linux/macos)"
+    "ENABLE_SANITIZER_UNDEFINED_BEHAVIOR\;OFF\;${SUPPORTS_UBSAN}\;Make certain types (numeric mostly) of undefined behavior into runtime errors"
+    "ENABLE_CPPCHECK\;OFF\;${MAKEFILE_OR_NINJA}\;Enable cppcheck analysis during compilation"
+    "ENABLE_IPO\;OFF\;OFF\;Enable whole-program optimization"
+    "ENABLE_INCLUDE_WHAT_YOU_USE\;OFF\;OFF\;Enable include-what-you-use analysis during compilation"
+    "ENABLE_PCH\;OFF\;OFF\;Enable pre-compiled-headers support"
```
aminya, line 66: Should we automate this and exclude mingw on Windows?

src/Utilities.cmake
```diff
@@ -82,3 +82,31 @@ function(set_env_from_string env_string)
...
+      PARENT_SCOPE)
+endfunction()
+
+macro(get_all_targets_recursive targets dir)
+  get_property(
+    subdirectories
+    DIRECTORY ${dir}
+    PROPERTY SUBDIRECTORIES)
+  foreach(subdir ${subdirectories})
+    get_all_targets_recursive(${targets} ${subdir})
```
ddalcino, line 101: Two questions:
1. Is the `get_all_targets` function used anywhere? Do we need it for this PR?
2. Maybe I don't really understand how macros work in CMake. According to the [docs](https://cmake.org/cmake/help/latest/command/macro.html#macro-vs-function), "A macro is executed as if the macro body were pasted in place of the calling statement." I would like my fears allayed that CMake won't just textually replace all instances of `get_all_targets_recursive` with the function definition, ala the c preprocessor, and get stuck in an infinite loop trying to process this. I would hope that the CMake devs wouldn't allow this, but still...

ddalcino (approved):
I think these changes to `DynamicProjectOptions.cmake` can only make this library more useful than it already is. I like the distinction between the 'user' and 'developer' default option sets. Well done.

src/DynamicProjectOptions.cmake
lefticus: I don't want to 100% exclude any compiler from any setting, in case things change in the near future, if that is what you mean. Just more like "if you enable something that's not the default, then you're on your own" was my philosophy here.

src/Utilities.cmake
lefticus: TBH I copy and pasted this from Stack Overflow, I didn't write it, but it does work.

It is specifically used here: https://github.com/cpp-best-practices/cpp_starter_project/pull/181/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR95-R100 as a method to get ASAN/Windows/Visual Studio IDE playing nicely together.

ddalcino: I remember trying code from stackoverflow when we were discussing #101. I tried printing out the whole list of targets that the function retrieved (on Windows), and there were the 4 targets defined in cpp_starter_project, alongside ~10 others that I didn’t recognize. Have you looked at the output of this function, and if so, are you ok with it?

src/DynamicProjectOptions.cmake
aminya: I meant setting the default like the Address sanitizer.

src/Utilities.cmake
lefticus: Yeah, it's working fine how it is, because it not only sets the PATH for regular executables for debugging, but also for test targets, and "building" the test targets inside of MSVC executes them correctly with the path to ASAN set up properly

src/DynamicProjectOptions.cmake
lefticus: Sorry, I'm not sure what you're asking. I did automate the defaults for things like ASAN on Windows, etc, up above this section:

https://github.com/cpp-best-practices/cpp_starter_project/pull/181/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR95-R100

aminya: I am saying that we can automate PCH's default like how we did for ASAN

```
    "ENABLE_PCH\;OFF\;${NOT_WIN32_AND_GNU}\;Enable pre-compiled-headers support"
```

aminya (approved):

aminya:
We can address my comment on the PCH header default in a separate PR
https://github.com/cpp-best-practices/project_options/pull/46#discussion_r796102393

src/DynamicProjectOptions.cmake
lefticus: Oh, OK, but I would reverse that. On for User, off for Developer. Just to speed up build times for the user and reduce the risk of accidentally including headers during development.

I went Off/Off ultimately since you had expressed some discomfort with it, and it was low priority to me.

PR #50, opened 2022-01-29
fix: bug fixes in run_vcpkg function

aminya:
Fixes some bugs in the automatic installation of vcpkg and also automatic updating of the repository.


aminya:
I have tested it, and it seems it works.

PR #56, opened 2022-02-04
fix: use the cmake's include install directory for target_include_system_libraries

### LINKED ISSUE #55: target_link_system_libraries's compatibility for header only installations
I faced a bug where using `target_link_system_libraries` caused an issue on a header-only target that was being installed using `install` function 

```cmake
target_link_system_libraries(header_only_target INTERFACE some_target)
```

aminya:
PUBLIC and INTERFACE mean that when the package is installed, the public/interface headers should be provided on the installed environment. This issue is not seen when a library is linked as private because during the installation, the headers are already built into the library.

Fixes #55

### DIFF
--- a/src/SystemLink.cmake
+++ b/src/SystemLink.cmake
@@ -10,15 +10,23 @@ function(target_include_system_directories target)
 
   foreach(scope IN ITEMS INTERFACE PUBLIC PRIVATE)
     foreach(lib_include_dirs IN LISTS ARG_${scope})
-      if(MSVC)
+      if(NOT MSVC)
         # system includes do not work in MSVC
         # awaiting https://gitlab.kitware.com/cmake/cmake/-/issues/18272#
         # awaiting https://gitlab.kitware.com/cmake/cmake/-/issues/17904
-        target_include_directories(${target} ${scope} ${lib_include_dirs})
+        set(_SYSTEM SYSTEM)
+      endif()
+      if(${scope} STREQUAL "INTERFACE" OR ${scope} STREQUAL "PUBLIC")
+        target_include_directories(
+          ${target}
+          ${_SYSTEM}
+          ${scope}
+          "$<BUILD_INTERFACE:${lib_include_dirs}>"
+          "$<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR}>")
       else()
         target_include_directories(
           ${target}
-          SYSTEM
+          ${_SYSTEM}
           ${scope}
           ${lib_include_dirs})
       endif()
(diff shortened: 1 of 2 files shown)


src/SystemLink.cmake
```diff
@@ -10,15 +10,23 @@ function(target_include_system_directories target)
...
-        target_include_directories(${target} ${scope} ${lib_include_dirs})
+        set(_SYSTEM SYSTEM)
+      endif()
+      if(scope STREQUAL INTERFACE)
+        target_include_directories(
+          ${target}
+          ${_SYSTEM}
+          ${scope}
+          "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/${lib_include_dirs}>"
+          "$<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR}>")
```
aminya, line 25: This seems to be needed to allow installing interface libraries. @lefticus Do you have a suggestion regarding this?

src/SystemLink.cmake
```diff
@@ -10,15 +10,23 @@ function(target_include_system_directories target)
...
     foreach(lib_include_dirs IN LISTS ARG_${scope})
-      if(MSVC)
+      if(NOT MSVC)
         # system includes do not work in MSVC
         # awaiting https://gitlab.kitware.com/cmake/cmake/-/issues/18272#
         # awaiting https://gitlab.kitware.com/cmake/cmake/-/issues/17904
-        target_include_directories(${target} ${scope} ${lib_include_dirs})
+        set(_SYSTEM SYSTEM)
+      endif()
+      if(scope STREQUAL INTERFACE)
```
aminya, line 19: Not sure why this only happens in an interface target

src/SystemLink.cmake
```diff
@@ -10,15 +10,23 @@ function(target_include_system_directories target)
...
     foreach(lib_include_dirs IN LISTS ARG_${scope})
-      if(MSVC)
+      if(NOT MSVC)
         # system includes do not work in MSVC
         # awaiting https://gitlab.kitware.com/cmake/cmake/-/issues/18272#
         # awaiting https://gitlab.kitware.com/cmake/cmake/-/issues/17904
-        target_include_directories(${target} ${scope} ${lib_include_dirs})
+        set(_SYSTEM SYSTEM)
+      endif()
+      if(${scope} STREQUAL "INTERFACE" OR ${scope} STREQUAL "PUBLIC")
```
aminya, line 19: Not sure why this bug only happens in an INTERFACE target and not PRIVATE

lefticus:
@aminya I'm trying to review this right now, and I am curious where this function target_include_system_directories, target_link_system_libraries is used?

aminya:
> @aminya I'm trying to review this right now, and I am curious where this function target_include_system_directories, target_link_system_libraries is used?

These two functions are not used in the library. They are two free functions that the users can use optionally to suppress the warnings of their dependencies. I personally just use this instead of `target_link_libraries` whenever I find a package with vcpkg or Conan.

src/SystemLink.cmake
aminya: I found the reason. It is because public and interface mean that when the package is installed they should be provided by the installer (not during build).

aminya:
I found the reason for the bug this fixes. It is because public and interface mean that when the package is installed, the public headers should be provided on the installed environment.

PR #58, opened 2022-02-10
feat: add package_project function

### LINKED ISSUE #52: Add `package_project` function to package a project for external usage from vcpkg, conan, etc
Manually packaging/exporting a project for external users via the built-in CMake functions like `install` and `configure_package_config_file` is cumbersome. This whole thing can be abstracted away using a simple function that does it all.

Some examples of such functions:

https://github.com/TheLartians/PackageProject.cmake
https://github.com/robotology/ycm/blob/master/modules/InstallBasicPackageFiles.cmake

Related to https://github.com/cpp-best-practices/cpp_starter_project/issues/78

aminya:
Fixes #52 

- [x] add documentation to readme
- [x] test this with direct `--install` command on Windows, Linux, and macOS with all the compilers
- [x] add CI to test installation
- [x]  add `find_package` test from another project
- [x] tested with executable
- [x] tested with header-only library
- [x] tested with a normal library

This adds `package_project` function to make it easy and automatic to package a project.

Not only this function installs the package, but also the generated package can be found using the following code from a different CMake project. Its dependencies (e.g. fmt and Eigen3) are also found automatically using `find_dependency`.

```cmake
find_package(myproject::mylibrary CONFIG REQUIRED)
target_link_libraries(another_target PRIVATE myproject::mylibrary)
```

To be able to make this work, it is required to generate several files such as `myprojectConfig.cmake`, etc. The other benefit of this is that the defaults are set for modern package managers like vcpkg and Conan.

For example, running `task test_install` generates the following. I added another project that uses the installed package and uses it as a test. This should show you the validity and capabilities of the code added.

image

Maybe later:

- [ ] test with vcpkg 
- [ ] test with Conan. 
- [ ] It would be great to generate a vcpkg port file or Conan recipe


src/PackageProject.cmake
```diff
@@ -0,0 +1,155 @@
+# Uses ycm (permissive BSD-3-Clause license) and FowardArguments (permissive MIT license)
+
+macro(package_project)
+  set(_options # default to true
+      ARCH_INDEPENDENT)
+  set(_oneValueArgs
+      # default all of these to the project_name or the given name:
+      NAME
```
lefticus, line 8: I think defaulting this to project_name could lead to confusing errors for future users. I find very rarely is a project simple enough that the project_name is also the exact name of the target you want to export.

aminya: I strongly believe that the name of the project should be the same as the name of the package. Otherwise, the project is just confusing everyone. For `TARGETS`, sometimes, multiple targets are needed to be exported. So, I think we can require the `TARGETS` property, but not the `NAME`. In an ideal world, the name of the project should be the namespace, prefix, etc, but the additional target names can be different.

Note that these options are related to [ycm's options](https://github.com/robotology/ycm/blob/1860bad63dcdf6727c1ce0746561e1cada0dccef/modules/InstallBasicPackageFiles.cmake#L17). I removed a lot of them that cause more pain than gains.

lefticus: I see, I thought NAME was the name of one given target.

aminya: Modified in 
https://github.com/cpp-best-practices/project_options/pull/58/commits/769883e641f331ba0af7f3a18dc163823057b767
and
https://github.com/cpp-best-practices/project_options/pull/58/commits/26f45ae5370543d558a6e8de1ddf8bd7115da117

src/PackageProject.cmake
```diff
@@ -0,0 +1,143 @@
...
+      PUBLIC_INCLUDES
+      # the names of the INTERFACE/PUBLIC dependencies that are found using `CONFIG`
+      PUBLIC_DEPENDENCIES_CONFIGED
+      # the INTERFACE/PUBLIC dependencies that are found by any means using `find_dependency`.
+      # the arguments must be specified within double quotes (e.g. "<dependency> 1.0.0 EXACT" or "<dependency> CONFIG").
+      PUBLIC_DEPENDENCIES
+      # the names of the PRIVATE dependencies that are found using `CONFIG`. Only included when BUILD_SHARED_LIBS is OFF.
+      PRIVATE_DEPENDENCIES_CONFIGED
+      # PRIVATE dependencies that are only included when BUILD_SHARED_LIBS is OFF
+      PRIVATE_DEPENDENCIES)
```
aminya, line 29: I wonder if these can be automatically found from the properties of the targets.

src/PackageProject.cmake
```diff
@@ -0,0 +1,157 @@
...
+      CONFIG_EXPORT_DESTINATION
+      # default to ${CMAKE_INSTALL_DATADIR}/cmake/${NAME} suitable for vcpkg, etc.
+      CONFIG_INSTALL_DESTINATION)
+  set(_multiValueArgs
+      # recursively found for the current folder if not specified
+      TARGETS
+      # a list of public/interface include directories or files
+      PUBLIC_INCLUDES
+      # the names of the INTERFACE/PUBLIC dependencies that are found using `CONFIG`
+      PUBLIC_DEPENDENCIES_CONFIGED
```
aminya, line 26: Should this be named `PUBLIC_DEPENDENCIES_BY_CONFIG` instead?

aminya: https://github.com/cpp-best-practices/project_options/pull/58/commits/816fd6b89899a759fae0392af111592dbbfe758b

ddalcino:
I want to approve this, but I don't have enough time to evaluate it yet.

It would help me a lot if we had some CI running here. If we had just a couple dummy projects building in a GitHub action that demonstrate the features this project offers, including packaging, it would be a lot easier to see that the changes are working.

lefticus:
> I want to approve this, but I don't have enough time to evaluate it yet.
> 
> It would help me a lot if we had some CI running here. If we had just a couple dummy projects building in a GitHub action that demonstrate the features this project offers, including packaging, it would be a lot easier to see that the changes are working.

This absolutely needs an example / test of some sort to see what it generates.

For example: I have packaging working with ~3 lines of CMake on all 3 platforms with one of my forks of `cpp_starter_project`, with posting to github releases via github actions. So I need to understand what specifically this gains.

aminya:
I have added three tests to this PR. 
- One project with an executable and a header-only library
- One project with an executable
- One project with a header-only library

If you run `task test` (see Taskfile.yaml`), it will generate the install files in the `./test/build` directory.

> For example: I have packaging working with ~3 lines of CMake on all 3 platforms with one of my forks of cpp_starter_project, with posting to github releases via github actions. So I need to understand what specifically this gains.

This does more than "just" installing a package.  The generated package can be found using the following code. Its dependencies (fmt and Eigen3) are also found automatically using `find_dependency`.

```cmake
find_package(myproject::mylibrary CONFIG REQUIRED)
target_link_libraries(another_target PRIVATE myproject::mylibrary)
```
To be able to make this work, it is required to generate several files such as `myprojectConfig.cmake`, etc. The other benefit of this is that the defaults are set for modern package managers like vcpkg.

I try to write another test that finds the package using the above code.

ddalcino:
> I have added three tests to this PR. 
> 
> - One project with an executable and a header-only library
> 
> - One project with an executable
> 
> - One project with a header-only library
> 
> 
> 
> If you run `task test` (see `Taskfile.yaml`), it will generate the install files in the `./test/build` directory.

Can we please have a GitHub action that does this?

aminya:
Here you. I have added the CI and the tests. 
Run the following to test it:
```
task test_release -- -DCMAKE_INSTALL_PREFIX="./install"
```

aminya:
I added another project that uses the installed package and uses it as a test. This should show you the validity and capabilities of the code added.

See `./test_install` folder.

You can test it by running the following (this also runs in the CI):

```
task test_install
```

aminya:
Weren't there any comments, objections, or approvals?

lefticus (approved):
It's a large PR, but it does seem that this is a valuable tool to add

aminya:
You're welcome. Most of these are various tests and documentation to support the feature.

PR #66, opened 2022-02-12
Set template so IDE/tooling picks up cppcheck warnings

lefticus:


src/StaticAnalyzers.cmake
```diff
@@ -2,11 +2,18 @@ macro(enable_cppcheck CPPCHECK_OPTIONS)
   find_program(CPPCHECK cppcheck)
   if(CPPCHECK)
 
+    if(CMAKE_GENERATOR MATCHES ".*Visual Studio.*")
```
aminya, line 5: Shouldn't this be `${CMAKE_GENERATOR}`. I find CMake's rules for variables still confusing.

aminya (approved):

PR #70, opened 2022-02-15
fix: correctly check if vsinstalldir is in $path on msvc

LtdSauce:
Before this commit having a falsly setup VSINSTALLDIR would not trigger the error. 
This had two reasons:
- first, the CMake string FIND module had the arguments mixed up
- second, the check itself never expanded the output variable from the string FIND

After this commit the error should be emitted if $VCINSTALLDIR is not in $PATH


LtdSauce:
Sadly, after this it is still possible that the check succeeds but the asan runtime lib is still not found. @aminya i am not sure if this check is enough to ensure valid asan support for msvc.

aminya (approved):
Thanks for catching the bug. Leaving the docs for the record 
https://cmake.org/cmake/help/latest/command/string.html?highlight=string#find

PR #71, opened 2022-02-17
fix: make Conan optional in dynamic_project_options

aminya:
Many people use vcpkg

BREAKING CHANGE this requires adding `set(ENABLE_CONAN_DEFAULT ON)` before calling `dynamic_project_options`

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,9 @@ project(myproject LANGUAGES CXX C)
 # Set PCH to be on by default for all non-Developer Mode Builds
 set(ENABLE_PCH_USER_DEFAULT ON)
 
+# enable Conan
+set(ENABLE_CONAN_DEFAULT ON)
+
 # Initialize project_options variable related to this project
 # This overwrites `project_options` and sets `project_warnings`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -299,6 +299,9 @@ project(myproject LANGUAGES CXX C)
 # Set PCH to be on by default for all non-Developer Mode Builds
 set(ENABLE_PCH_USER_DEFAULT ON)
 
+# enable Conan
+set(ENABLE_CONAN_DEFAULT ON)
```
aminya, line 303: @lefticus is this the preferred method for setting options for `dynamic_project_options`? Can't this be an argument passed to the function?

ljishen: I think ENABLE_* should be done before the include of `DynamicProjectOptions.cmake`. Because the defaults are updated before calling `dynamic_project_options()`.

aminya: In my opinion, this is not a modular design. The defaults should be called once the function is called, not when it is included.

aminya: Created an issue for it
https://github.com/cpp-best-practices/project_options/issues/72

PR #76, opened 2022-02-19
fix: remove non-C warnings for C

### LINKED ISSUE #69: Warning is valid for C++/ObjC++ but not for C
When compiling with GCC, it complains some warnings are not for C. We may need to set different warnings for different languages.

```
cc1: warning: command-line option ‘-Wnon-virtual-dtor’ is valid for C++/ObjC++ but not for C
cc1: warning: command-line option ‘-Wold-style-cast’ is valid for C++/ObjC++ but not for C                                                                                            
cc1: warning: command-line option ‘-Woverloaded-virtual’ is valid for C++/ObjC++ but not for C
cc1: warning: command-line option ‘-Wuseless-cast’ is valid for C++/ObjC++ but not for C
```

aminya:
Fixes #69

### DIFF
--- a/src/CompilerWarnings.cmake
+++ b/src/CompilerWarnings.cmake
@@ -99,8 +99,15 @@ function(
     # TODO support Intel compiler
   endif()
 
-  # use the same warning flags for C
+  # Add C warnings
   set(PROJECT_WARNINGS_C "${PROJECT_WARNINGS_CXX}")
+  list(
+    REMOVE_ITEM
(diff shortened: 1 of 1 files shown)


aminya:
Not sure if this is all we need to exclude for C. But #69 only specified these.

lefticus (approved):

ljishen:
This looks good to fix the problem. I only see these warnings being complained about in my program.

Issue #64, opened 2022-02-11
Enable StaticAnalyzers for C


ljishen:
Currently only CXX are supported by cppcheck, clang-tidy, and include-what-you-use. It would be great to have them enabled for C as well.

- [x] Set C standard: #65 
- [ ] Clang-tidy for C #78
- [ ] Cppecheck for C #77

aminya:
Also, #78 for clang-tidy

```
https://github.com/cpp-best-practices/project_options/archive/refs/heads/c-clangtidy.zip
```

ljishen:
@aminya Yes, I'll let you know.

ljishen:
@aminya I've posted the feedback on the page of the corresponding pull request.

PR #77, opened 2022-02-19
feat: add cppcheck for C

aminya:
#64


ljishen:
This works except that the support of C std in cppcheck is only up to c11: https://github.com/danmar/cppcheck/blob/main/cli/cmdlineparser.cpp#L772-L773

However, this project can support up to c17: https://github.com/cpp-best-practices/project_options/blob/main/src/Standards.cmake#L26

If c17 is set by the user, then the following error will occur:
```
cppcheck: error: unrecognized command line option: "--std=c17".
```

Adding a check of the `CMAKE_C_STANDARD` can help solve this problem.

aminya:
I opened an issue for cppcheck so they address the lack of support for C17.
https://sourceforge.net/p/cppcheck/discussion/general/thread/19ea152bba/

I also disabled setting C17 when cppcheck is enabled.

ljishen:
Cool, thanks!

PR #78, opened 2022-02-19
feat: add clang-tidy for C

aminya:
#64


ljishen:
This change works as expected. Thanks!

PR #79, opened 2022-02-19
feat: make Conan quiet if cmake is not verbose

### LINKED ISSUE #12: If CMake is quiet, Conan should be quiet too
The verbosity levels should be passed to Conan as is in CMake.

aminya:
Fixes #12


src/Conan.cmake
```diff
@@ -32,6 +32,11 @@ macro(run_conan)
     set(LIST_OF_BUILD_TYPES ${CMAKE_CONFIGURATION_TYPES})
   endif()
 
+  is_verbose(_is_verbose)
+  if(NOT ${_is_verbose})
+    set(CONAN_QUIET YES)
+  endif()
```
ddalcino, line 38: I am unclear about the consequences of this when `_is_verbose` is true. As far as I can tell, the CONAN_QUIET variable will just be an empty string when you try to use it. Should it not be set to “NO” or “OFF” or something?

lefticus: Agreed, it does seem like we need something different here

aminya: Fixed in https://github.com/cpp-best-practices/project_options/pull/79/commits/fa2266813748dcdbd4dd335d112b0344333eca27

aminya:
@lefticus @ddalcino please test this locally on your system. For me, Conan is so fragile that I cannot make it work.

aminya:
The tests show that this is working as expected.

PR #80, opened 2022-02-19
fix: move the options of dynamic_project_options into itself

### LINKED ISSUE #72: Set the defaults of dynamic_project_options inside the function itself
The defaults should be called once the function is called, not when it is included.

_Originally posted by @aminya in https://github.com/cpp-best-practices/project_options/pull/71#discussion_r810441485_

aminya:
Fixes #72


ddalcino:
This sounds like a reasonable idea, but GitHub really mangled the diff here.

It looks, to me, like you just moved the `macro(dynamic_project_options)` line up 120 lines, and indented everything in between. If that’s what this is, then I approve. However, I would need to read it closely line by line to verify that. The diff makes it look like an enormous change.

Are there any other changes here that I’m overlooking?

aminya:
There is a way to hide the whitespace changes:

image

ddalcino:
D’oh! How did I not know that? Thanks for letting me know!

ddalcino (approved):

PR #83, opened 2022-02-19
feat: add enable_native_optimizations + rename IPO option

aminya:
Closes https://github.com/cpp-best-practices/cpp_starter_project/pull/105


ddalcino (approved):
This looks straightforward enough, and I can’t see any drawbacks to this. Could be really useful.

PR #88, opened 2022-03-01
feat: upgrade to doxygen-awesome-css v2.0.2

aminya:
Here are the release notes. I don't see any breaking changes that affect `project_options`.
https://github.com/jothepro/doxygen-awesome-css/releases/tag/v2.0.0

Thanks to @jothepro for making Doxygen awesome.

How it looks like for our tests:
Web capture_2-3-2022_3598_

### DIFF
--- a/src/Doxygen.cmake
+++ b/src/Doxygen.cmake
@@ -29,9 +29,9 @@ function(enable_doxygen DOXYGEN_THEME)
 
   if("${DOXYGEN_THEME}" STREQUAL "awesome" OR "${DOXYGEN_THEME}" STREQUAL "awesome-sidebar")
     # use a modern doxygen theme
-    # https://github.com/jothepro/doxygen-awesome-css v2.0.0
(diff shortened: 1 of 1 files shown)


jothepro:
Afaik to inject Javascript into Doxygen you have to provide a header template, like described in the install instructions of the extensions. If there would be an easier way I'd love to know about it as well! 

So I think the answer is no, there is no easy way to automate the addition of the extensions with just the CMake Doxygen module. Maybe you could ship a custom header template with your CMake module, with all the extensions installed? But personally as a user I'd not be a fan of that bc I need control over the header template for injecting SEO-Metadata and the "[github corner](https://github.com/tholman/github-corners)".

It may be worth a try to directly ask your question in [doyxgen/doxygen](https://github.com/doxygen/doxygen)?

ddalcino (approved):

aminya:
Thanks for the answer! I will create an issue so we can look into this later.

PR #91, opened 2022-03-02
fix: use WARNINGS_AS_ERRORS for pch + clang-tidy error

aminya:


aminya:
PCH doesn't work nicely with GCC for some reason.

aminya:
Oh, wow. Another CMake surprise. Using `return()` inside Macro causes the other script to return! How can we return from a macro then?

aminya:
OK. I manged to use `else()` instead of `return()`

aminya:
It seems that PCH on GCC fails when `cmake --install` is used in another package.

aminya:
I think we should disable PCH once clang-tidy is enabled for a non-clang compiler. It is more important to run clang-tidy (semantic check) than using PCH (only a compile-time hack).

PR #92, opened 2022-03-02
fix: use a warning if LTO is not supported

aminya:
If an optimization is missing on the toolchain, it does not mean that the build should fail. This discourages enabling this option


aminya:
MacOS LTO seems to be failing with LLVM enabled.

aminya:
This is fixed. The reason was that the `test_install` project didn't have LTO enabled, and so it could not link with the libraries of the other project compiled with LTO. I added a TODO for setting this option in `package_project` for all the users of the project that has LTO enabled.

PR #98, opened 2022-03-06
fix: enable native optimization only on x64

### LINKED ISSUE #96: Arm Clang/GCC do not support '-march=native'
This flag fails on some ARM platforms including the Apple M1 ([1](https://github.com/DMOJ/judge-server/issues/303), [2](https://stackoverflow.com/questions/65966969/why-does-march-native-not-work-on-apple-m1)). We may be able to check whether it is supported by using `clang -march=native -cx /dev/null` ([1](https://stackoverflow.com/questions/52557417/how-to-check-support-compile-flag-in-autoconf-for-clang), [2](https://stackoverflow.com/questions/2658461/what-predefined-macro-can-i-use-to-detect-clang)) before adding to the compile options.

aminya:
Fixes #96


ljishen:
Would you like to use the trial-and-error approach so that you won't sacrifice the possibility of applying the optimization?

aminya:
If there is a deterministic way to do this, I will be open to doing that.

Gcc has a separate set of options for different architectures. 
https://gcc.gnu.org/onlinedocs/gcc/Submodel-Options.html#Submodel-Options

It specifies `-march=native` in the arm options section, but it didn't work inside Godbolt.
https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#ARM-Options

aminya:
For Clang, I see the code for it here: Not sure why it is not enabled though
https://github.com/llvm/llvm-project/blob/17a68065c378da74805e4e1b9a5b78cc9f83e580/clang/lib/Driver/ToolChains/Arch/ARM.cpp#L893

aminya:
For M1 Mac, it is under review:
https://reviews.llvm.org/D119788

ljishen:
> If there is a deterministic and cross-platform way to do this, I will be open to doing that.
> 
> GCC has a separate set of options for different architectures. https://gcc.gnu.org/onlinedocs/gcc/Submodel-Options.html#Submodel-Options
> 
> It specifies `-march=native` in the arm options section, but it didn't work inside Godbolt. https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#ARM-Options

I think a cross-platform solution without user input is almost impossible. A deterministic solution is also hard given the continuous (micro)architecture innovations for different CPUs, both GCC and clang are just keep incorporating the new changes. Therefore, I think a trial-and-error approach may be a better option for guiding in this situation, ie, test the flag before applying. I've tested on my ARM machine with the `-mcpu=native` on GCC and clang, both of them work.

As a relevant suggestion, we may set up an ARM-based environment using the [Oracle A1 Instances](https://www.servethehome.com/oracle-cloud-giving-away-ampere-arm-a1-instances-always-free/) (which says "Always Free") for better testing this project.

aminya:
Yeah, if a trial and error solution is available I am open to adding it.

aminya:
Changed #96 to reflect this.

ljishen:
> Yeah, if a trial and error solution is available I am open to adding it.

For GCC and clang, we can use `[gcc/clang] -Werror [the test flag] -E -xc /dev/null` and check the return code.
For example, `[gcc/clang] -Werror -march=native -E -xc /dev/null` or `[gcc/clang] -Werror -mcpu=native -E -xc /dev/null`.

I guess MSVC will have a similar way to test, but I don't have the environment for testing.

PR #115, opened 2022-03-27
Allow Conan profiles

### LINKED ISSUE #114: Allow the user to use a conan profile
Sometimes it can be challenging to convince Conan to use the settings that you want it to. Conan allows users to use a [profile](https://docs.conan.io/en/latest/reference/profiles.html) to provide fine grain control over what settings Conan uses. See [Conan instructions for using Emscripten](https://docs.conan.io/en/latest/integrations/cross_platform/emscripten.html) for a common use case for Conan profiles.

Currently, `project_options` forces the user to use [`conan_cmake_autodetect()`](https://github.com/conan-io/cmake-conan#conan_cmake_autodetect) to determine settings automatically. A Conan profile, if desired, should replace this step, and allow the user to choose all of these settings explicitly.

**What I would like:** I think that a user should be allowed to pass a Conan profile as an argument to the `project_options` function. When that variable is not empty, `project_options` should skip `conan_cmake_autodetect()`, and call [`conan_cmake_install()`](https://github.com/conan-io/cmake-conan#conan_cmake_install) with the argument `PROFILE ${PATH_TO_CONAN_PROFILE}`.

ddalcino:
Fix #114 

Todo:
- [ ] Add a test for this feature. Currently, I am in the process of trying to get a cpp_starter_project to build properly using emscripten, using a branch of `project_options` that implements this feature (unsuccessfully so far). I can tell you that this branch of `project_options` passes the profile properly into Conan, but I would prefer to prove it with a test.


aminya:
@ddalcino any update on this?

ddalcino:
I think this merge is right; it's hard to tell without good tests.

aminya (approved):

aminya:
The low-level test you are talking about should be done in conan-cmake. We are just calling it with the argument they defined. 

We can add a simple Conan profile in CI and test with that just for verification.

PR #105, opened 2022-03-17
include missing GNUInstallDirs

ClausKlein:
add symlink to Index.cmake to enable use with CPM.cmake module too

This fix https://github.com/cpp-best-practices/project_options/issues/104 too

PR #116, opened 2022-03-28
feat: add CMakeLists.txt to support CPM

aminya:
Co-Authored-By: Claus Klein <1833050+ClausKlein@users.noreply.github.com>

Cherry-picked from #105 

- [ ] test that Conan works in the projects that use project_options

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,3 @@
+include_guard()
+
+include("${CMAKE_CURRENT_LIST_DIR}/src/Index.cmake")


ClausKlein:
**Great**.
By the way, xhy do you not use a symbolic link to `Index.cmake?`

aminya:
> **Great**.
> By the way, xhy do you not use a symbolic link to `Index.cmake?`
> 

Because symlinks on Windows require admin permission

aminya:
@ddalcino Could you test this change with Conan? I don't have many Conan projects to test this with

ddalcino:
> @ddalcino Could you test this change with Conan? I don't have many Conan projects to test this with

I must be missing some context here. I'm not sure what you're asking for. I can test this manually with [CMake Package Manager](https://github.com/cpm-cmake/CPM.cmake) if that will help.

Are you asking me to create a Conan package that uses project_options, and see if I can use it in a dependent project? That's probably a good idea, but I think it should be tested continuously rather than as a one-off.

ClausKlein:
> @ddalcino Could you test this change with Conan? I don't have many Conan projects to test this with

I tested under ubuntu this: https://github.com/ClausKlein/cpp_starter_project/commit/d690b81ec76351218bf8b08d52651fd2ceb3f474

ddalcino:
I have tested this, and I can confirm that it works as intended. CI logs are here: https://github.com/ddalcino/cpp_starter_project/actions/runs/2072971834

For completeness, I have also verified that CPM does not load `project_options` properly without this change.

I think it could be really useful to add directions to the README about how to fetch `project_options` properly using CPM; I didn't know how to do it myself.

ddalcino (approved):

aminya:
Thanks for the testing!

PR #119, opened 2022-04-04
Allow list of files instead of Doxygen theme

dk949:
Allows users to specify a list of custom theme files which will be added
to `DOXYGEN_HTML_EXTRA_STYLESHEET`.

If any of the files are not found, the theme is reverted to "original".
It reverts to "original" and not "awesome", because (presumably) the
user still wishes to use the custom theme and just mistyped the path, so
there is no need to download the `awesome` theme.

Note: In the `project_options` macro the `DOXYGEN_THEME` flag was made
to take multiple parameters for this to work.


src/Doxygen.cmake
```diff
@@ -42,8 +42,22 @@ function(enable_doxygen DOXYGEN_THEME)
       set(DOXYGEN_HTML_EXTRA_STYLESHEET ${DOXYGEN_HTML_EXTRA_STYLESHEET}
                                         "${_doxygen_theme_SOURCE_DIR}/doxygen-awesome-sidebar-only.css")
     endif()
-  else()
+  elseif("${DOXYGEN_THEME}" STREQUAL "original")
     # use the original doxygen theme
+  else()
+    # use custom doxygen theme
+
+    # if any of the custom theme files are not found, the theme is reveretd to original
```
aminya, line 50:
```suggestion
    # if any of the custom theme files are not found, the theme is reverted to original
```

aminya (changes requested):

aminya (approved):
Thanks for your contribution!

Issue #133, opened 2022-06-25
Static analyzer: suppress warning for 3rd-party/system libs


abeimler:
Is it possible to exclude targets or 3rd-party libs from the static analyzer ?

**Root Project**
```cmake
option(ENABLE_TESTING "Enable the tests" ${PROJECT_IS_TOP_LEVEL})
if(ENABLE_TESTING)
  set(ENABLE_CLANG_TIDY "ENABLE_CLANG_TIDY")
  set(ENABLE_CPPCHECK "ENABLE_CPPCHECK")
  set(ENABLE_COVERAGE "ENABLE_COVERAGE")

  if(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
    set(ENABLE_SANITIZER_ADDRESS "ENABLE_SANITIZER_ADDRESS")
    set(ENABLE_SANITIZER_UNDEFINED_BEHAVIOR "ENABLE_SANITIZER_UNDEFINED_BEHAVIOR")
  else()
    # or it is MSVC and has run vcvarsall
    string(FIND "$ENV{PATH}" "$ENV{VSINSTALLDIR}" index_of_vs_install_dir)
    if(MSVC AND "${index_of_vs_install_dir}" STREQUAL "-1")
      set(ENABLE_SANITIZER_ADDRESS "ENABLE_SANITIZER_ADDRESS")
    endif()
  endif()
endif()

# defaulted_project_options sets recommended defaults and provides user and developer
# modes and full GUI support for choosing options at configure time
# for more flexibility, look into project_options() macro
# Initialize project_options variable related to this project
# This overwrites `project_options` and sets `project_warnings`
# uncomment the options to enable them:
dynamic_project_options(
  ENABLE_CACHE
  ${ENABLE_CPPCHECK}
  ${ENABLE_CLANG_TIDY}
  ENABLE_VS_ANALYSIS
  ${ENABLE_COVERAGE}
  ${ENABLE_SANITIZER_ADDRESS}
  ${ENABLE_SANITIZER_UNDEFINED_BEHAVIOR}
  # Note: PCH is disabled by default in developer mode because these headers become
  # globally included and they can mask other errors
  PCH_HEADERS
  # This is a list of headers to pre-compile, here are some common ones
  <vector>
  <string>
  <utility>
  <array>
  <algorithm>
  <concepts>
  CPPCHECK_OPTIONS
  --enable=style,performance,warning,portability
  --inline-suppr
[...]

Issue #132, opened 2022-06-15
Disable linting/checks on specific target


ryanseipp:
The project as a whole is using v0.22.4 and calls `dynamic_project_options` with `ENABLE_CONAN` and `CONAN_OPTIONS` parameters. CMake is generating a debug build and developer mode is on.

I have a library within my project that utilizes protobuf/grpc and generates a lot of code. This library never links against `project_options` or `project_warnings`. That library is then linked to the executable that requires it using `target_link_system_libraries(exe PRIVATE protoObjects)`. However, I still get tons of errors against that generated code. The sanitizers/linters would be great to use across the rest of the app, but not this single target. As such, I'm hoping there's an option, workaround, or otherwise to be able to prevent static analysis against this target.

ryanseipp:
I've found the solution to this. Didn't realize at first that this was functionality and properties that CMake enables by default when a target is created after `CMAKE_CXX_CPPCHECK` is defined. The solution is to unset the `CXX_CPPCHECK` and `CXX_CLANG_TIDY` properties on the target using
```cmake
set_property(TARGET protoObjects PROPERTY CXX_CPPCHECK)
set_property(TARGET protoObjects PROPERTY CXX_CLANG_TIDY)
```

There may be an opportunity to improve this experience by exporting a function to disable static analysis on a target, but as my issue has a solution, I'll close this.

aminya:
Thanks! Would you like to add your comment to the readme?

PR #134, opened 2022-06-27
feat: add disable static analysis macros (#133)

abeimler:
* add macro `target_disable_clang_tidy`
* add macro `target_disable_cpp_check`
* add macro `target_disable_static_analysis`

#133 #132

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -110,3 +110,4 @@ words:
   - "shlib"
   - "vcpkg"
   - cppdbg
+  - mythirdpartylib


aminya (approved):

abeimler:
> Some of the windows configurations fail because of clang-tidy.

Don't know if this is the best solution, couldn't test it fully on my windows (the project isn't set up fully with conan, clang-tidy, doxygen is missing etc.)

abeimler:
idk if `Eigen` is really a good example ibh ...
Maybe using something like `test/src/mylib2` is a better test, or make a new `test/libs/mythird-party-lib` with warnings triggered.

header-only libs should be ignored by using it as a system library `target_link_system_libraries`

aminya (approved):
Thanks for the work!

PR #137, opened 2022-07-15
feat: add VCPKG_REV option for run_vcpkg

aminya:
This option allows checking out a specific branch name or a commit sha.
If `VCPKG_REV` is set to a specific commit sha, the builds will become reproducible as that exact commit is always used for the builds.


ljishen:
I like this feature. Thanks!

aminya:
> I like this feature. Thanks!

You're welcome! 🙂

PR #139, opened 2022-07-21
fix: typo in mingw.cmake

Kushal-Chandar:
changes copmiler to compiler

### DIFF
--- a/src/MinGW.cmake
+++ b/src/MinGW.cmake
@@ -18,7 +18,7 @@ function(is_mingw value)
     return()
   endif()
 
-  # if the copmiler is unknown by CMake
+  # if the compiler is unknown by CMake
   if(NOT CMAKE_CXX_COMPILER
      AND NOT CMAKE_C_COMPILER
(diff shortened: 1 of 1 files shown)


aminya:
I wonder why this was not caught by our spell checker tests.

Kushal-Chandar:
> I wonder why this was not caught by our spell checker tests.

May be the word has to be added to [cspell.config.yaml](https://github.com/aminya/project_options/blob/main/cspell.config.yaml) but i am not sure.
[This check](https://github.com/aminya/project_options/runs/7455142411?check_suite_focus=true) failed as doxygen was not installed, is this an issue with the check or did I do something wrong ?
I edited this [file](https://github.com/aminya/project_options/blob/main/src/MinGW.cmake) directly in github without cloning the repo. Does this have anything to do with the failing check ?

aminya:
The CI issue is not because of this PR. SourceForge has become unreliable lately. 
https://github.com/aminya/setup-cpp/issues/101

aminya (approved):

PR #150, opened 2022-09-05
feat: support multi-project projects by adding PREFIX

### LINKED ISSUE #149: Error trying to have multiple projects that use project_options
There's a decent chance this is pure user error and I don't understand how to compose project_options properly, but I'm getting configure-time errors when trying to have multiple projects using project_options, specifically:

```
[cmake] CMake Error at build/_deps/_project_options-src/src/Index.cmake:125 (add_library):
[cmake]   add_library cannot create target "project_options" because another target
[cmake]   with the same name already exists.  The existing target is an interface
[cmake]   library created in source directory
[cmake]   "/home/mcoding/Projects/project_options_multiple_include_mwe".  See
[cmake]   documentation for policy CMP0002 for more details.
```

The project structure is like this:

```
project/
    apps/
        CMakeLists.txt
        main.cpp
    CMakeLists.txt
    external/mylib/
        CMakeLists.txt
        include/mylib
            mylib.hpp
```

Both `project` and the submodule `mylib` use `project_options`, which explains why cmake thinks `project_options` is being multiply defined, because both `project` and `mylib` do indeed define their own set of project options. Is this a situation where `project_options` should be defining target names that depend on the project name like `${PROJECT_NAME}_project_options` instead of a global name `project_options`? Or is there some way to have `project_options` scoped to a namespace like `mylib::project_options`? Or should only the top-level project be defining project options (using an `if TARGET` guard before including it) with subprojects modifying the options using `dynamic_project_options`?

Any guidance here would be appreciated.

Here are the relevant list files for the minimal working example:

Top level `CMakeLists.txt`:
```
cmake_minimum_required(VERSION 3.16)

include(FetchContent)
FetchContent_Declare(_project_options URL https://github.com/aminya/project_options/archive/refs/tags/v0.24.1.zip)
FetchContent_MakeAvailable(_project_options)
include(${_project_options_SOURCE_DIR}/Index.cmake)
include(${_project_options_SOURCE_DIR}/src/DynamicProjectOptions.cmake)

project(myproject LANGUAGES CXX C)

dynamic_project_options()

add_subdirectory(external/mylib)

add_subdirectory(apps)
```

`apps/CMakeLists.txt`
```
add_executable(main main.cpp)
target_link_libraries(main PRIVATE project_options project_warnings)

target_link_system_libraries(
  main
  PRIVATE mylib
)
```

`external/mylib/CMakeLists.txt`
```
cmake_minimum_required(VERSION 3.16)

include(FetchContent)
FetchContent_Declare(_project_options URL https://github.com/aminya/project_options/archive/refs/tags/v0.24.1.zip)
FetchContent_MakeAvailable(_project_options)
include(${_project_options_SOURCE_DIR}/Index.cmake)
include(${_project_options_SOURCE_DIR}/src/DynamicProjectOptions.cmake)

project(mylib LANGUAGES CXX C)

dynamic_project_options()

add_library(mylib INTERFACE)
target_link_libraries(mylib INTERFACE project_options project_warnings)

set(INCLUDE_DIR "include") 
target_include_directories(mylib INTERFACE "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/${INCLUDE_DIR}>"
                                          "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
```

aminya:
Fixes #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ It accepts the following named flags:
 
 It gets the following named parameters that can have different values in front of them:
 
+- `PREFIX`: the optional prefix that is used to define `${PREFIX}_project_options` and `${PREFIX}_project_warnings` targets when the function is used in a multi-project fashion.
 - `DOXYGEN_THEME`: the name of the Doxygen theme to use. Supported themes:
   - `awesome-sidebar` (default)
   - `awesome`
--- a/src/Index.cmake
+++ b/src/Index.cmake
@@ -41,6 +41,7 @@ include("${ProjectOptions_SRC_DIR}/Vcpkg.cmake")
 
 #
 # Params:
+# - PREFIX: the optional prefix to be prepended to the `project_options` and `project_warnings` targets when the function is used in a multi-project fashion.
 # - WARNINGS_AS_ERRORS: Treat compiler warnings as errors
 # - ENABLE_CPPCHECK: Enable static analysis with cppcheck
 # - ENABLE_CLANG_TIDY: Enable static analysis with clang-tidy
@@ -94,7 +95,11 @@ macro(project_options)
       ENABLE_SANITIZER_UNDEFINED_BEHAVIOR
       ENABLE_SANITIZER_THREAD
       ENABLE_SANITIZER_MEMORY)
-  set(oneValueArgs LINKER VS_ANALYSIS_RULESET CONAN_PROFILE)
+  set(oneValueArgs
+      PREFIX
+      LINKER
+      VS_ANALYSIS_RULESET
+      CONAN_PROFILE)
   set(multiValueArgs
       DOXYGEN_THEME
       MSVC_WARNINGS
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -221,6 +221,7 @@ It accepts the following named flags:
 
 It gets the following named parameters that can have different values in front of them:
 
+- `PREFIX`: the optional prefix that is used to define `${PREFIX}_project_options` and `${PREFIX}_project_warnings` targets when the function is used in a multi-project fashion.
```
ClausKlein, line 224: **IMHO**: For may reasons, this should **NOT** optional -> **it must be mandatory!**

1. konsider: project A is installed (with exported `project_options`) and later, project B, which use also  `project_options`) depends from project A. It use find_package() and import  `project_options`
2. Or Project D depends from both: A and C, but does not use  `project_options`. But A and C use independent.

One solution for this future desaster is to never export  `project_options` and  `project_warnings`!
They must be used `PRVATE `only and never exported.

ClausKlein: **It is not a good idea to export static-analyser settings or gcov options!**

aminya: The default example only enables the static analyzers when the `FEATURE_TESTS ` is enabled

https://github.com/aminya/project_options#usage

ClausKlein: OK, I allways disable install in developer mode!

aminya: Do not turn off the installation. Just put the test and analyzer under an option. This is very common even without this library

aminya: I don't know which package manager you use, but the install rules are required for vcpkg and conan to work without any issues.

Issue #61, opened 2022-02-10
Adding more compiler warnings by default


aminya:
There are other Clang warnings that we have not added. I think these will be beneficial to add

```
-Wformat-overflow
-Wformat-nonliteral
-Wformat-truncation=2
-Wformat-security
-fno-common
-Wcast-qual
-Wdeprecated
-Wmissing-declarations
-Wmissing-prototypes
-Wnested-externs
-Wold-style-definition
-Wpointer-arith
-Wsign-compare
-Wstrict-prototypes
-Wundef
-Wwrite-strings
```

lefticus:
I fully agree that we should do a pass over the existing warning flags set, but I think it should be more nuanced than this.

Right now I know there is some redundant flags. For example, the `-Wwrite-strings` is listed as "enabled by default" on Clang's warning page. I know there are others that are currently enabled that we should consider removing while we add new ones that were not previously covered.

Another one to consider is `-pedantic-errors`, which according to the GCC documentation does "This is not equivalent to -Werror=pedantic, since there are errors enabled by this option and not enabled by the latter and vice versa." but it gives no further explanation.

ljishen:
Some of these warnings may only be available for C and not C++ in GCC. For example, `-Wmissing-prototypes`, `-Wnested-externs`, `-Wold-style-definition`, and `-Wstrict-prototypes`.

janwilmans:
more suggestions: `-Wconversion -Wsign-conversion -Wmissing-include-dirs -Wformat=2 -Wunused -Wunused-variable -Wcast-align -Wno-vla`

janwilmans:
and (but only for C++) `-Wnon-virtual-dtor -Woverloaded-virtual`

ClausKlein:
'-Wno-vla' seems to disable a warning?

aminya:
> more suggestions:
> 
> ```
> -Wconversion 
> -Wsign-conversion 
> -Wmissing-include-dirs 
> -Wformat=2 
> -Wunused 
> -Wunused-variable 
> -Wcast-align 
> -Wno-vla 
> -Wnull-dereference
> -Wshadow
> ```

Many of these are already enabled. Could you find the ones that are not?
https://github.com/aminya/project_options/blob/main/src/CompilerWarnings.cmake

aminya:
> '-Wno-vla'

It disallows the use of variable-length C arrays. I am wondering if this is already covered by -Wpedantic

[...]

PR #151, opened 2022-09-11
Warn about semicolon after in-class function definition

ClausKlein:
#61


aminya (approved):
Thanks! Feel free to add more warnings in the future (#61)

PR #153, opened 2022-09-14
Forcing Conan build type to match CMake one

mous16:
Invoking Conan with `-s build_type=` parameter matching active CMake build type, or all the active ones if in multi-config scenario.

This allows CMake to correctly build targets and link to correct Conan-insatlled dependencies.

### DIFF
--- a/src/Conan.cmake
+++ b/src/Conan.cmake
@@ -69,7 +69,7 @@ macro(run_conan)
         set(CONAN_ENV ENV "CC=${CMAKE_C_COMPILER}" "CXX=${CMAKE_CXX_COMPILER}")
       else()
         # Derive all conan settings from a conan profile
-        set(CONAN_SETTINGS PROFILE ${ProjectOptions_CONAN_PROFILE})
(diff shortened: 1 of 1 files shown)


mous16:
> Is this desirable? I usually prefer to build the dependencies as Release unless there is an issue I need to debug in them.

I thought the same, until I stumbled across MSVC, which builds iterators with some additional introspection features when in Debug mode, making them not binary compatible with Release mode builded ones.  [Here](https://docs.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170) more infos abaout.
So, tring to link a debug executable with a non debug library results in errors.

Moreover, using Ninja in multi-config mode, together with Conan with `cmake_find_package_multi` generator, causes CMake to generate only `<MODULE>-release.cmake` file, making it impossible to build in Debug mode (nor in RelWithDebInfo or MinSizeRel).

I'll try to upload a minimal example as soon as possible.

aminya (approved):
Thanks for the clarification!

aminya:
Feel free to add the example as a test.

PR #163, opened 2022-10-08
ProjectOptions_SRC_DIR is now cached as INTERNAL

### LINKED ISSUE #161: Can no longer include project_options as a submodule
I prefer to explicitly depend on project_options as a submodule and add it using `add_directory`. Unfortunately, this seems to have broken with #154 as I now get the following error:

```
[cmake] CMake Error at cmake/project_options/src/Common.cmake:5 (include):
[cmake]   include could not find requested file:
[cmake] 
[cmake]     /PreventInSourceBuilds.cmake
```

SK83RJOSH:
Fixes #161

### DIFF
--- a/src/Common.cmake
+++ b/src/Common.cmake
@@ -1,7 +1,17 @@
 include_guard()
 
+# This function sets ProjectOptions_SRC_DIR using the current lists path
+macro(set_project_options_src_dir)
+  get_directory_property(LISTFILE_STACK LISTFILE_STACK)
+  list(POP_BACK LISTFILE_STACK _LIST_FILE)
+  cmake_path(GET _LIST_FILE PARENT_PATH ProjectOptions_SRC_DIR)
+endmacro()
+
 # Common project settings run by default for all the projects that call `project_options()`
 macro(common_project_options)
+  set_project_options_src_dir()
+  message(DEBUG "${ProjectOptions_SRC_DIR}")
+
   include("${ProjectOptions_SRC_DIR}/PreventInSourceBuilds.cmake")
   assure_out_of_source_builds()
 


src/Index.cmake
```diff
@@ -11,7 +11,7 @@ endif()
 include_guard()
 
 # only useable here
-set(ProjectOptions_SRC_DIR ${CMAKE_CURRENT_LIST_DIR})
+set(ProjectOptions_SRC_DIR ${CMAKE_CURRENT_LIST_DIR} CACHE INTERNAL "")
```
ClausKlein, line 14: This must not a `CMakeCache` variable because if more than one subproject use this ProjectOptions independently, there is a name clash!

aminya: We should probably add a note about this. If someone uses project options in multiple submodules, it can result in errors

SK83RJOSH: The issue we're running into is that this variable is seemingly out of scope by the time a lot of project_option macros are called. The only solution I can think of would be to call the includes before macro definitions – but I'm not sure what the impact on build times would be since I guess the inclusions are delayed intentionally?

ClausKlein: > The issue we're running into is that this variable is seemingly out of scope by the time a lot of project_option macros are called. The only solution I can think of would be to call the includes before macro definitions – but I'm not sure what the impact on build times would be since I guess the inclusions are delayed intentionally?

you may try to change the [project_options](https://github.com/aminya/project_options)/CMakeLists.txt and append the `CMAKE_MODULE_PATH` ...

src/Common.cmake
```diff
@@ -1,7 +1,17 @@
 include_guard()
 
+# This function sets ProjectOptions_SRC_DIR using the current lists path
+macro(set_project_options_src_dir)
+  get_directory_property(LISTFILE_STACK LISTFILE_STACK)
+  list(POP_BACK LISTFILE_STACK _LIST_FILE)
+  cmake_path(GET _LIST_FILE PARENT_PATH ProjectOptions_SRC_DIR)
```
ClausKlein, line 7: sorry, this is new in `cmake version v3.20`!

@aminya should we change min required version?

see too [cmake_minimum_required(VERSION 3.20)](https://github.com/ClausKlein/WindowsToolchain/blob/7fde47cd3dabd5a5b11b00a646a9b51d850a1a5f/Windows.Clang.toolchain.cmake#L50)

aminya: Can you rewrite this without cmake_path? This feature is not worth the version upgrade.

ClausKlein: > Can you rewrite this without cmake_path? This feature is not worth the version upgrade.

but the `Toolchain` requirers this too?

SK83RJOSH: I would also like to say that 3.20 was released over a year and a half ago, on March 23rd, 2021. How conservative do we need to be? Debian stable targets 3.18, and that's about as conservative as it gets... :)

ClausKlein: > I would also like to say that 3.20 was released over a year and a half ago, on March 23rd, 2021. How conservative do we need to be? Debian stable targets 3.18 (July, 2020), and that's about as conservative as it gets... :)

... and with `pip install cmake ninja ` cmake v3.24 may be available.

aminya: Yeah, I was following the Debian Stable version. That's why we are using 3.18 now. But if the Windows toolchain requires that, we can update the required version.

aminya: Fixed in #166

aminya (approved):

PR #174, opened 2022-11-29
Make sure cmake min and max version don't change

### LINKED ISSUE #173: cmake policy warnings
Hi. I've upgraded to a new VS build tools which has CMake 3.24 and started getting the following warning
```
[cmake] CMake Warning (dev) at C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.24/Modules/FetchContent.cmake:1264 (message):
[cmake]   The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
[cmake]   not set.  The policy's OLD behavior will be used.  When using a URL
[cmake]   download, the timestamps of extracted files should preferably be that of
[cmake]   the time of extraction, otherwise code that depends on the extracted
[cmake]   contents might not be rebuilt if the URL changes.  The OLD behavior
[cmake]   preserves the timestamps from the archive instead, but this is usually not
[cmake]   what you want.  Update your project to the NEW behavior or specify the
[cmake]   DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
[cmake]   robustness issue.
[cmake] Call Stack (most recent call first):
[cmake]   out/build/windows-msvc-debug-user-mode-user/_deps/_project_options-src/src/PackageProject.cmake:160 (FetchContent_Declare)
[cmake]   CMakeLists.txt:175 (package_project)
[cmake] This warning is for project developers.  Use -Wno-dev to suppress it.
```

I tried increasing the maximum version in my main CMakeLists.txt
```
cmake_minimum_required(VERSION 3.15...3.24)
```

but the warning kept showing. Running cmake with `--trace` showed that `project_options` calls again to `cmake_minimum_required` at [`Index.cmake`](https://github.com/aminya/project_options/blob/ffb1b59a59299a106d2cb07dd0bd9f6dbee8fc4b/src/Index.cmake#L1) which seems to override whatever the project declares. Removing that line the warning was no longer displayed.

I think it's better for the library to not call `cmake_minimum_required` itself given the above.

dvirtz:
Fixes #173

### DIFF
--- a/src/Index.cmake
+++ b/src/Index.cmake
@@ -1,7 +1,3 @@
-cmake_minimum_required(VERSION 3.20)
-# 3.20 is required by the windows toolchain and cmake_path. It also has a more reliable building functionality.
-# 3.18 required by package_project and interprocedural optimization. It also has a more reliable building functionality (no errors during the linking stage).
-
 include_guard()
 
 # only useable here
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,4 +1,6 @@
-cmake_minimum_required(VERSION 3.16...3.21)
+set(CMAKE_MIN_VERSION 3.16)
+set(CMAKE_MAX_VERSION 3.21)
+cmake_minimum_required(VERSION ${CMAKE_MIN_VERSION}...${CMAKE_MAX_VERSION})
 
 # set a default CXX standard used by the external tools like clang-tidy, cppcheck, etc.
 # You can later set fine-grained standards for each target using `target_compile_features`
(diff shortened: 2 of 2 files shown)


src/Index.cmake
```diff
@@ -1,7 +1,3 @@
-cmake_minimum_required(VERSION 3.20)
```
ClausKlein, line 1: this is the real min version! Do not change.

test/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
-cmake_minimum_required(VERSION 3.16...3.21)
```
ClausKlein, line 1: should be:
`cmake_minimum_required(VERSION 3.20...3.25)`

test/CMakeLists.txt
```diff
@@ -11,6 +13,18 @@ set(CMAKE_CXX_STANDARD 20)
...
+if(NOT CMAKE_MINIMUM_REQUIRED_VERSION STREQUAL ${CMAKE_MIN_VERSION})
+  message(FATAL_ERROR "CMake minimum version changed to ${CMAKE_MINIMUM_REQUIRED_VERSION}")
+endif()
+cmake_policy(GET CMP0126 CMP126)
+cmake_policy(GET CMP0127 CMP127)
+if(NOT CMP126 STREQUAL NEW OR NOT CMP127 STREQUAL "")
+  message(FATAL_ERROR "CMake policy settings don't match maximum version ${CMAKE_MAX_VERSION} (CMP126=${CMP126} CMP127=${CMP127})")
+endif()
+
+
```
ClausKlein, line 27: This seems me too much?

This is for intern tests only.

src/Index.cmake
aminya: project_options doesn't work with older versions. Please revret

aminya, line 1: project_options doesn't work with older versions. Please revert

aminya (changes requested):

PR #178, opened 2022-12-21
Update Doxygen.cmake - Fixes #177

### LINKED ISSUE #177: Doxygen options overwrite
In the CMake for Doxygen there is a block that overwrites custom settings:

https://github.com/aminya/project_options/blob/283c25fdd782f2f3e754a09488175e910893b180/src/Doxygen.cmake#L15-L21

If I don't want call graphs or caller graphs, there is currently no way to disable it since it always gets overwritten by those values.

Can you please add a defined if statement around it so if it is already set, that project_options does not touch the set setting?

Subtixx:
Check if the settings we set are already set by the user, and if so don't overwrite them.

This fixes the just opened issue #177


aminya (approved):
Thanks for the fix

Subtixx:
🤣 The failing windows test is doxygen again.. 🤣

aminya:
Haha. I think I fixed the download issue by having a retry mechanism in setup-cpp. 
https://github.com/aminya/setup-cpp/pull/155

PR #182, opened 2022-12-28
feat: add more optimization flags (no exception/no rtti)

abeimler:
Just to make life easier with MSVC- vs. GNU-flags.


src/Index.cmake
```diff
@@ -85,6 +85,8 @@ macro(project_options)
       ENABLE_DOXYGEN
       ENABLE_INTERPROCEDURAL_OPTIMIZATION
       ENABLE_NATIVE_OPTIMIZATION
+      DISABLE_EXCPETIONS
```
glennvl, line 88:
```suggestion
      DISABLE_EXCEPTIONS
```

src/Index.cmake
```diff
@@ -163,6 +165,13 @@ macro(project_options)
     enable_native_optimization(${_options_target})
   endif()
 
+  if(${ProjectOptions_DISABLE_EXCPETIONS})
```
glennvl, line 168:
```suggestion
  if(${ProjectOptions_DISABLE_EXCEPTIONS})
```

src/Optimization.cmake
```diff
@@ -32,3 +32,13 @@ macro(enable_native_optimization project_name)
     endif()
   endif()
 endmacro()
+
+macro(disable_exceptions project_name)
+  target_compile_options(${project_name} INTERFACE $<$<CXX_COMPILER_ID:MSVC>:/EHs-c- /D_HAS_EXCEPTIONS=0>)
+  target_compile_options(${project_name} INTERFACE $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-fno-exceptions>)
```
glennvl, line 38: Is it enough to add only `-fno-exceptions` or should `-fno-unwind-tables` also be added?

src/DynamicProjectOptions.cmake
```diff
@@ -78,6 +78,8 @@ macro(dynamic_project_options)
       "ENABLE_CPPCHECK\;OFF\;${MAKEFILE_OR_NINJA}\;Enable cppcheck analysis during compilation"
       "ENABLE_INTERPROCEDURAL_OPTIMIZATION\;OFF\;OFF\;Enable whole-program optimization (e.g. LTO)"
       "ENABLE_NATIVE_OPTIMIZATION\;OFF\;OFF\;Enable the optimizations specific to the build machine (e.g. SSE4_1, AVX2, etc.)."
+      "DISABLE_EXCEPTIONS\;OFF\;OFF\;Disable Exceptions (no-exceptions flag)"
+      "DISABLE_RTTI\;OFF\;OFF\;Disable RTTI (no-rtti flag)"
```
glennvl, line 82: These are interesting options for embedded targets. There's a third one that comes to mind `-fno-threadsafe-statics`.

src/Optimization.cmake
abeimler: Seems fine for me ... https://alex-robenko.gitbook.io/bare_metal_cpp/compiler_output/exceptions

_(I personally use this flags for some embedded and [ETL](https://www.etlcpp.com/) stuff :P)_

src/DynamicProjectOptions.cmake
glennvl: The reason I mentioned that flag is because it may not be desirable (or possible at all) to have mutexes (or some other form of critical sections / disabling interrupts) when static variables are accessed on some systems. A quick google tells that MSVC has a `/Zc:threadSafeInit-` flag.

I took an interests in your pull request because I'm trying to do similar stuff. @aminya is probably in a better position to judge.

aminya: Since this optimization is unsafe, I prefer that we do not add it. In general, It would be important to not hinder the safety/correctness of a program for minor performance improvements.

aminya: If you are sure that one part of the code doesn't need to be thread-safe, you should use the thread-local attributes to specify that specific variable. You cannot assume that this is true for all the programs and your dependencies.
https://en.wikipedia.org/wiki/Thread-local_storage#C_and_C++

aminya (approved):

aminya (approved):

PR #183, opened 2022-12-30
test: move tests and break the Taskfile files + refactoring

abeimler:
Just try out some refactoring ... project gets bigger and bigger ...

* move test_../ into sub-folder tests/
* split Taskfile.yml into sub tasks
* use [Styleguide](https://taskfile.dev/styleguide/) for Taskfile
* add more examples ...
* WIP: windows tasks


tests/.gitignore
```diff
@@ -0,0 +1,767 @@
+!install/
```
aminya, line 1: This file can be simplified significantly. For example, we don't need the Node stuff.

aminya (changes requested):

aminya (changes requested):
Should enable submodules inside the CI files

```
      - uses: actions/checkout@v3
        with:
          submodules: true
  ```

File: tests/minimal/Taskfile.yml [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+  build:mingw:from-env:
+    env: 
+      CC: x86_64-w64-mingw32-gcc
+      CXX: x86_64-w64-mingw32-g++
+    cmds:
+      - task: build
+        vars:
+          CMAKE_ARGS: -DENABLE_CROSS_COMPILING:BOOL=ON
+
+  build:mingw:from-triplet:
```
aminya, line 27: I prefer `.` as a name separator and not `:`

.github/workflows/ci.yml
```diff
@@ -82,9 +84,9 @@ jobs:
 
       - name: Test
         run: |
-          task test
-          task test_release
-          task test_install
+          task main:test
+          task main:test:release
+          task install:test
```
aminya, line 89: `install:test` doesn't exist.

aminya (changes requested):

tests/minimal/Taskfile.yml
abeimler: Any other name conventions or style changes ?

_Would be nice to write down some Guidelines into CONTRIBUTING.md, some other differences from [Taskfile Styleguide](https://taskfile.dev/styleguide/)._  
_I think everything else can be linted._

**Note:**
> Use colon for task namespacing
>
> This is also done automatically when using included Taskfiles.

https://taskfile.dev/styleguide/#use-colon-for-task-namespacing

_Not sure if this breaks the style a bit when using includes: `minimal:build.mingw.from-triplet` vs `minimal:build:mingw:from-triplet`_

aminya: Yeah, it breaks the specified style, but I prefer using `.`. To me using `:` is kind of strange. I haven't seen any programming language using that for accessing children.

aminya (approved):

tests/minimal/Taskfile.yml
aminya: It seems that `:` here means the sub-tasks of a Taskfile. I wish they had used `.` for that too.

PR #186, opened 2023-01-05
Fix  #185 - make conancenter's remote name to match conan's default

### LINKED ISSUE #185: conancenter remote name is not conan's default
nitially reported in https://github.com/aminya/project_options/issues/148#issuecomment-1372303229 :

To me the problem is really that it uses a different name that the conan default, so basically it renames `conancenter` to `cci` for 99% of users.

https://github.com/aminya/project_options/blob/53eb453c5918c94e9e5041fba55b2f7362ddc293/src/Conan.cmake#L23-L29

This is the conan default: 

https://github.com/conan-io/conan/blob/c3fc425871149b12a1d12153bed90a06a3d20132/conans/client/cache/remote_registry.py#L14

```python
CONAN_CENTER_REMOTE_NAME = "conancenter"
```

jmarrec:
- Fix  #185

### DIFF
--- a/src/Conan.cmake
+++ b/src/Conan.cmake
@@ -19,10 +19,10 @@ macro(run_conan)
   include(${CMAKE_BINARY_DIR}/conan.cmake)
 
   # Add (or remove) remotes as needed
-  # conan_add_remote(NAME conan-center URL https://conan.bintray.com)
+  # conan_add_remote(NAME conan-center URL https://center.conan.io)
   conan_add_remote(
     NAME
-    cci
+    conancenter
     URL
     https://center.conan.io
     INDEX


src/Conan.cmake
```diff
@@ -19,10 +19,10 @@ macro(run_conan)
   include(${CMAKE_BINARY_DIR}/conan.cmake)
 
   # Add (or remove) remotes as needed
-  # conan_add_remote(NAME conan-center URL https://conan.bintray.com)
+  # conan_add_remote(NAME conan-center URL https://center.conan.io)
   conan_add_remote(
     NAME
-    cci
+    conancenter
```
jmarrec, line 25: It uses a different name that the conan default, so basically it renames `conancenter` to `cci` for 99% of users.

This is the conan default: 

https://github.com/conan-io/conan/blob/c3fc425871149b12a1d12153bed90a06a3d20132/conans/client/cache/remote_registry.py#L14

```python
CONAN_CENTER_REMOTE_NAME = "conancenter"
```

aminya (approved):

PR #190, opened 2023-01-19
feat: create a symlink of `complie_commands.json` on the source dir

FeignClaims:
It helps clang based tools to find `compile_commands.json` when using out-of-source builds, and to find the correct `compile_commands.json` when handling multiple builds.
Already tested in my [FeignClaims/cmake_starter_template](https://github.com/FeignClaims/cmake_starter_template/tree/main/cmake) (for the orign extension version, see `cmake/SymlinkCompileCommands.cmake`) when I switch between clang and gcc on MacOS and Windows.

Users should be notified to add `compile_commands.json` to `.gitignore` though.

### DIFF
--- a/src/Common.cmake
+++ b/src/Common.cmake
@@ -99,13 +99,13 @@ macro(common_project_options)
       file(CREATE_LINK ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json SYMBOLIC)
     endif()
 
-    # Add `compile_commans.json` to `.gitignore` if `.gitignore` exists
+    # Add compile_commans.json to .gitignore if .gitignore exists
     set(GITIGNORE_FILE "${CMAKE_SOURCE_DIR}/.gitignore")
     if(EXISTS ${GITIGNORE_FILE})
       file(STRINGS ${GITIGNORE_FILE} HAS_IGNORED REGEX "^compile_commands.json")
 
       if(NOT HAS_IGNORED)
-        message(STATUS "Adding `compile_commands.json` to `.gitignore`")
+        message(STATUS "Adding compile_commands.json to .gitignore")
         file(APPEND ${GITIGNORE_FILE} "\ncompile_commands.json")
       endif()
     endif()
--- a/src/Common.cmake
+++ b/src/Common.cmake
@@ -76,12 +76,9 @@ macro(common_project_options)
     # Make a symbol link of compile_commands.json on the source dir to help clang based tools find it
     if(WIN32)
       # Detect whether cmake is run as administrator (only administrator can read the LOCAL SERVICE account reg key)
-      cmake_host_system_information(RESULT IS_ADMINISTRATOR QUERY WINDOWS_REGISTRY "HKU\\S-1-5-19")
+      cmake_host_system_information(RESULT QURY_RESULT QUERY WINDOWS_REGISTRY "HKU\\S-1-5-19" ERROR_VARIABLE IS_NONADMINISTRATOR)
 
-      if(IS_ADMINISTRATOR)
-        # For administrator, symlink is available
-        file(CREATE_LINK ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json SYMBOLIC)
-      else()
+      if(IS_NONADMINISTRATOR)
         # For non-administrator, create an auxiliary target and ask user to run it
         add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/compile_commands.json
           COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json
@@ -93,6 +90,9 @@ macro(common_project_options)
           VERBATIM
         )
         message(AUTHOR_WARNING "Can't symlink compile_commands.json. Run cmake as administrator or use `cmake --build <build_dir> -t _copy_compile_commands` to copy it.")
+      else()
+        # For administrator, symlink is available
+        file(CREATE_LINK ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json SYMBOLIC)
       endif()
     else()
       file(CREATE_LINK ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json SYMBOLIC)
--- a/src/Common.cmake
+++ b/src/Common.cmake
@@ -86,14 +86,14 @@ macro(common_project_options)
 
     if(IS_NONADMINISTRATOR)
       # For non-administrator, create an auxiliary target and ask user to run it
-      add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/compile_commands.json
-        COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json
-        DEPENDS ${CMAKE_SOURCE_DIR}/compile_commands.json
-        VERBATIM
+      add_custom_command(OUTPUT ${CMAKE_SOURCE_DIR}/compile_commands.json
+          COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json
+          DEPENDS ${CMAKE_BINARY_DIR}/compile_commands.json
+          VERBATIM
       )
       add_custom_target(_copy_compile_commands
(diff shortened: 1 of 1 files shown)


src/Common.cmake
```diff
@@ -69,8 +69,15 @@ macro(common_project_options)
...
   endif()
 
-  # Generate compile_commands.json to make it easier to work with clang based tools
-  set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+  if(CMAKE_GENERATOR MATCHES ".*Makefile*." OR CMAKE_GENERATOR MATCHES ".*Ninja*")
+    # Generate compile_commands.json to make it easier to work with clang based tools
+    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+    # Make a symbol link of compile_commands.json on the source dir to help clang based tools find it
+    execute_process(
```
aminya, line 77: On Windows creating symlinks requires admin access, so this command will fail. Could you detect that and use copy instead?

aminya (changes requested):

aminya:
> Users should be notified to add compile_commands.json to .gitignore though.

It would be great if we could automate that!

src/Common.cmake
FeignClaims: This cannot be done using copy because cmake generates `compile_commands.json` at the very end of generation stage - that's why I use symlink.

I tried `file(GENERATE)` to copy it but the command executed at the beginning of generation stage.

The final possible way is to use `add_custom_command` and `add_custom_target`. However, this adds an annoying target to the user's view and requires user to run it on one's own.

aminya: Another possibility is to detect admin access (potentially by shelling out). If access is not provided, simlinking can be skipped and a command for the manual copying can be added.

src/Common.cmake
```diff
@@ -69,8 +69,35 @@ macro(common_project_options)
...
+        add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/compile_commands.json
+          COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json
+          DEPENDS ${CMAKE_SOURCE_DIR}/compile_commands.json
+          VERBATIM
+        )
+        add_custom_target(_copy_compile_commands
+          DEPENDS ${CMAKE_BINARY_DIR}/compile_commands.json
+          VERBATIM
+        )
+        message(AUTHOR_WARNING "Can't symlink compile_commands.json. Run cmake as administrator or use `cmake --build <build_dir> -t _copy_compile_commands` to copy it.")
```
aminya, line 95: Since this feature is optional, I would prefer a non-warning message.
```suggestion
        message(STATUS "compile_commands was not symlinked to the root. Run `cmake --build <build_dir> -t _copy_compile_commands` if needed.")
```

aminya (changes requested):

src/Common.cmake
```diff
@@ -69,8 +69,35 @@ macro(common_project_options)
...
-  # Generate compile_commands.json to make it easier to work with clang based tools
-  set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+  if(CMAKE_GENERATOR MATCHES ".*Makefile*." OR CMAKE_GENERATOR MATCHES ".*Ninja*")
+    # Generate compile_commands.json to make it easier to work with clang based tools
+    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
+    # Make a symbol link of compile_commands.json on the source dir to help clang based tools find it
+    if(WIN32)
+      # Detect whether cmake is run as administrator (only administrator can read the LOCAL SERVICE account reg key)
+      cmake_host_system_information(RESULT IS_ADMINISTRATOR QUERY WINDOWS_REGISTRY "HKU\\S-1-5-19")
```
aminya: Nice job here 👍🏻

src/Common.cmake
FeignClaims: Done

src/Common.cmake
FeignClaims: Though quite an assumption. I can't test whether it works until I return home on February 18.

FeignClaims:
> > Users should be notified to add compile_commands.json to .gitignore though.
> 
> It would be great if we could automate the addition to `.gitignore`!

This should be optional then. I prefer to automately add it only if `.gitignore` exists.

aminya:
I tried this locally, but I got this error:
```
CMake Error at D:/GitHub/Cpp/project_options/src/Common.cmake:80 (cmake_host_system_information):
  cmake_host_system_information does not recognize <key> WINDOWS_REGISTRY
Call Stack (most recent call first):
  D:/GitHub/Cpp/project_options/src/Index.cmake:128 (common_project_options)
  CMakeLists.txt:53 (project_options)


CMake Error at D:/GitHub/Cpp/project_options/src/Common.cmake:96 (file):
  file failed to create symbolic link
  'D:/GitHub/Cpp/project_options/tests/myproj/compile_commands.json': A
  required privilege is not held by the client.

Call Stack (most recent call first):
  D:/GitHub/Cpp/project_options/src/Index.cmake:128 (common_project_options)
  CMakeLists.txt:53 (project_options)

```

FeignClaims:
> I tried this locally, but I got this error:
> 
> ```
> CMake Error at D:/GitHub/Cpp/project_options/src/Common.cmake:80 (cmake_host_system_information):
>   cmake_host_system_information does not recognize <key> WINDOWS_REGISTRY
> Call Stack (most recent call first):
>   D:/GitHub/Cpp/project_options/src/Index.cmake:128 (common_project_options)
>   CMakeLists.txt:53 (project_options)
> 
> 
> CMake Error at D:/GitHub/Cpp/project_options/src/Common.cmake:96 (file):
>   file failed to create symbolic link
>   'D:/GitHub/Cpp/project_options/tests/myproj/compile_commands.json': A
>   required privilege is not held by the client.
> 
> Call Stack (most recent call first):
>   D:/GitHub/Cpp/project_options/src/Index.cmake:128 (common_project_options)
>   CMakeLists.txt:53 (project_options)
> ```

I forgot the command requires CMake 3.24+. I'll try to call `reg.exe` directly using `execute_process`.

FeignClaims:
Should work now. Already tested in my windows virtual machine.

FeignClaims:
Here's even more question: I notice chances are that users already configured multiple cmake configurations (say gcc, clang). So when they switch between these configurations, they may not need to reconfigure it.

However, in the current implementation, I only give the non-administrator windows user the flexibility to copy `compile_commands.json` after configuration. Maybe this should be allowed for all users?

Here's a summay feature of my pr:

- if `.gitignore` exists in `${CMAKE_SOURCE_DIR}` and dosen't contain `compile_commands.json`, automatically add `compile_commands.json` into it. (this helps git based project.)
- If cmake generates `compile_commands.json` and user dosen't run cmake as non-administrator on Windows, symlink `${CMAKE_BUILD_DIR}/compile_commands.json` to `${CMAKE_SOURCE_DIR}/compile_commands.json`. (this helps clang based tools.)
- If cmake generates `compile_commands.json`, Add a custom command `_copy_compile_commands` to symlink/copy `${CMAKE_BUILD_DIR}/compile_commands.json` to `${CMAKE_SOURCE_DIR}/compile_commands.json`. (this helps users switch between different configurations without reconfiguration.)

FeignClaims:
Ready to merge now.

FeignClaims:
ready to merge now.

FeignClaims:
rebased on main and ready to merge now.

src/Common.cmake
```diff
@@ -69,8 +69,52 @@ macro(common_project_options)
...
+          VERBATIM
+      )
+      add_custom_target(_copy_compile_commands
+          DEPENDS ${CMAKE_SOURCE_DIR}/compile_commands.json
+          VERBATIM
+      )
+      message(STATUS "compile_commands.json was not symlinked to the root. Run `cmake --build <build_dir> -t _copy_compile_commands` if needed.")
+    else()
+      file(CREATE_LINK ${CMAKE_BINARY_DIR}/compile_commands.json ${CMAKE_SOURCE_DIR}/compile_commands.json SYMBOLIC)
+      message(STATUS "compile_commands.json was symlinked to the root.")
```
aminya, line 103:
```suggestion
      message(TRACE "compile_commands.json was symlinked to the root.")
```

src/Common.cmake
```diff
@@ -69,8 +69,52 @@ macro(common_project_options)
...
+    endif()
+
+    # Add compile_commans.json to .gitignore if .gitignore exists
+    set(GITIGNORE_FILE "${CMAKE_SOURCE_DIR}/.gitignore")
+
+    if(EXISTS ${GITIGNORE_FILE})
+      file(STRINGS ${GITIGNORE_FILE} HAS_IGNORED REGEX "^compile_commands.json")
+
+      if(NOT HAS_IGNORED)
+        message(STATUS "Adding compile_commands.json to .gitignore")
```
aminya, line 113:
```suggestion
        message(TRACE "Adding compile_commands.json to .gitignore")
```

aminya (commented):
Let's use `TRACE` for the messages that are not necessary for the user.

PR #192, opened 2023-01-19
fix: do not use the exception and RTTI optimizations for C

aminya:


aminya:
@abeimler Could you take a look into this?

PR #193, opened 2023-01-20
fix: cross arm C build

abeimler:
> fatal error: stdio.h: No such file or directory


aminya (approved):

aminya:
Strangely the error is back. @abeimler

PR #194, opened 2023-01-20
Added riscv architecture to Utilities.cmake

bwhitchurch:
Added statements to detect when system processor is either riscv64 or riscv32  and put rv64 or rv32 respectively,  into the arch variable. This keeps these cases from falling through and accidentally picking up the x64 arch flags.


aminya (approved):
Looks good to me. It would be great to have CI for RISKV.

PR #195, opened 2023-01-20
Allow passing build and host profiles through to conan

bwhitchurch:
Added additional variables with sensible defaults so that users may specify a host and/or build profile for the cmake_install_command. This is useful in cross compiling scenarios where the single profile argument is not enough.

### DIFF
--- a/src/Conan.cmake
+++ b/src/Conan.cmake
@@ -77,6 +77,23 @@ macro(run_conan)
         # CONAN_ENV should be redundant, since the profile can set CC & CXX
       endif()
 
+      if("${ProjectOptions_CONAN_PROFILE}" STREQUAL "")
+        set(CONAN_DEFAULT_PROFILE "default")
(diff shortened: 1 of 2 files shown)


src/Conan.cmake
```diff
@@ -77,6 +77,23 @@ macro(run_conan)
         # CONAN_ENV should be redundant, since the profile can set CC & CXX
       endif()
 
+      if("${ProjectOptions_CONAN_PROFILE}" STREQUAL "")
+        set(CONAN_DEFAULT_PROFILE "default")
+      else()
+        set(CONAN_DEFAULT_PROFILE ${ProjectOptions_CONAN_PROFILE})
+      endif()
```
aminya, line 84: This can be moved under the `if` section of the next condition. It would potentially simplify the code a bit.

aminya (approved):

PR #201, opened 2023-02-09
Add target disable macro for include-what-you-use

bogdan-lab:
It looks like for the interface completeness there should be macro `target_disable_include_what_you_use`.
Since `iwyu` is also static analysis tool, I added call for this macro in `target_disable_static_analysis`


aminya (approved):

bogdan-lab:
It seems that the test failed because the building process could not connect with git and it cannot be caused by changes introduced in this PR.

Do you expect me to solve this connection problem here?

aminya:
I guess it is fine. Thanks for the contribution

PR #199, opened 2023-01-31
feat: helper functions to help `package_project` find include directory and dependencies

FeignClaims:
In the current implementation, users set an independent variable to help `package_project` find include directory and dependencies, which is not preferred by modern cmake. What about differences among PRIVATE/PUBLIC/INTERFACE dependencies? Must I set these variables myself?

I'd like to add two functions - `target_include_header_directory` and `target_configure_dependencies` - to reduce the tedious work.

The reason why add `target_configure_dependencies` instead of extending `target_link_system_libraries` is that chances are that the package names and target names are often inconsistent and not logically deducible. (e.g. `tl-expected` vs. `tl::expected` but `Microsoft.GSL` vs. `Microsoft.GSL::GSL`.)

Before:

```cmake
add_library(my_lib "./src/my_lib/lib.cpp")
target_link_libraries(my_lib PRIVATE project_options project_warnings) # link project_options/warnings

# Includes
set(INCLUDE_DIR "include") # must be relative paths
target_include_directories(my_lib PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/${INCLUDE_DIR}>"
                                         "$<INSTALL_INTERFACE:./${CMAKE_INSTALL_INCLUDEDIR}>")

# Find dependencies:
set(my_lib_PUBLIC_DEPENDENCIES fmt)
set(my_lib_PRIVATE_DEPENDENCIES Eigen3)

foreach(DEPENDENCY ${my_lib_PUBLIC_DEPENDENCIES} ${my_lib_PRIVATE_DEPENDENCIES})
  find_package(${DEPENDENCY} CONFIG REQUIRED)
endforeach()

# Link dependencies:
target_link_system_libraries(my_lib
  PUBLIC
  fmt::fmt
  PRIVATE
  Eigen3::Eigen
)

# Package the project
package_project(
  TARGETS my_lib
  PUBLIC_DEPENDENCIES_CONFIGURED ${my_lib_PUBLIC_DEPENDENCIES}
  PUBLIC_INCLUDES ${INCLUDE_DIR}
)
```

After:

```cmake
add_library(my_lib "./src/my_lib/lib.cpp")
target_link_libraries(my_lib PRIVATE project_options project_warnings) # link project_options/warnings

# Includes
target_include_header_directory(my_lib)

# Find dependencies:
target_configure_dependencies(my_lib
  PUBLIC
  fmt
  PRIVATE
  Eigen3
)

# Link dependencies:
[...]

PR #202, opened 2023-03-29
feat: target functions to include interface directory and find dependencies

FeignClaims:
I assume this is a better version of #199. Instead of use global variables that are forced to change every time (that is, an incomplete simulation of global variables), I change to use target properties. 

- `target_include_interface_directories`: A wrapper that includes the header interface directory (usually named `include`).

  ```cmake
  target_include_interface_directories(<target_name> [<include_dir> ...])
  ```

- `target_find_dependencies`: A function that runs correct `find_package()` or `find_dependency()` on dependencies and binds the dependencies to a target.

  ```cmake
  target_find_dependencies(<target_name>
    # dependencies found by `find_dependency`
    [INTERFACE [dependency ...]]
    [PUBLIC [dependency ...]]
    [PRIVATE [dependency ...]]
    # dependencies found by `find_package(<dependency_name> CONFIG REQUIRED)`
    [INTERFACE_CONFIG [dependency ...]]
    [PUBLIC_CONFIG [dependency ...]]
    [PRIVATE_CONFIG [dependency ...]]
  )
  ```

The two functions can be called on the same target multiple times just like the builtin `target_link_libraries`.

In addition, after the extesion, the `package_project` can parse these properties automatically. Users are not required to pass includes and dependencies of targets manually. That is,

```cmake
add_library(my_lib "./src/my_lib/lib.cpp")
target_link_libraries(my_lib PRIVATE project_options project_warnings) # link project_options/warnings

# Includes
target_include_interface_directories(my_lib "${CMAKE_CURRENT_SOURCE_DIR}/include")

# Find dependencies:
target_find_dependencies(my_lib
  PRIVATE_CONFIG
  fmt
  Eigen3
)

# Link dependencies:
target_link_system_libraries(my_lib
  PRIVATE
  fmt::fmt
  Eigen3::Eigen
)

# Package the project
package_project(
  # Note that you must export `project_options` and `project_warnings` for `my_lib`
  TARGETS my_lib project_options project_warnings
)
```


aminya (dismissed):
Thanks for your contribution! 🎉 Looks great

PR #204, opened 2023-04-01
fix: `target_find_dependencies` should be a macro

FeignClaims:
When updating the corresponding project example to use #202 in aminya/cpp_vcpkg_project#19, I found a bug in current implementation.

`find_package` may bring some global "variables". For instance, `find_package(Catch2)` makes `include(Catch)` possible. In current implementation, however, `target_find_dependencies(... Catch2)` then `include(Catch)` may not work in some cases.

The wrapper `target_find_dependencies` should be a macro, just like the `find_dependency` in `include(CMakeFindDependencyMacro)`.


aminya (approved):
Thanks!

Issue #111, opened 2022-03-23
Overriding dynamic_project_options on CMake command line does not work


JesusKrists:
Trying to run a command such as:
`cmake -S . -B ./build -G "Ninja Multi-Config" -DCMAKE_BUILD_TYPE:STRING=Debug -DENABLE_DEVELOPER_MODE:BOOL=OFF -DOPT_ENABLE_COVERAGE:BOOL=ON -DOPT_ENABLE_INCLUDE_WHAT_YOU_USE:BOOL=OFF`

Results in the options not being overriden.
The CMakeCache.txt contains
```
//Analyze and report on coverage
OPT_ENABLE_COVERAGE:INTERNAL=ON
//Enable include-what-you-use analysis during compilation
OPT_ENABLE_INCLUDE_WHAT_YOU_USE:INTERNAL=OFF
```

Which seems correct, but my compile_commands.json does not contain any coverage flags and digging further into this I found that when DynamicProjectOptions.cmake:130 is parsed
```
cmake_dependent_option(
      OPT_${option_name}
      "${option_description}"
      ${option_developer_default}
      ENABLE_DEVELOPER_MODE
      ${option_user_default})
```
it overrides the command line option by the developer default or user default. In ENABLE_COVERAGE case it disables coverage in both configurations no matter what I pass to the command line

So this if here
```
if(OPT_${option_name})
  set(${option_name}_VALUE ${option_name})
else()
  unset(${option_name}_VALUE)
endif()
```
always unsets the value

aminya:
This is related to 	`dynamic_project_options`, which is maintained by @lefticus

aminya:
Fixed in #209

PR #208, opened 2023-04-06
New feature: allow handling of project_options parameters through CMake options

mous16:
## Rational
Many of the parameters I usually pass to `project_options()` depend on build-specific traits, and I often need to change them from one build to another.  
This situation leads me to manually create CMake options via `option()` to be able to configure the build differently from time to time.

## Proposed solution
I've been using this approach on my projects for some time now, and I think it might be useful to integrate it into project_option.
With this changes, the user, instead of calling `project_options()` in his CMakeLists.txt passing all the required information, can call `setup_project_options()` passing only the prefix to be used for target generation.  
`setup_project_options()` generates all required `option()` calls, and takes care of calling `project_options()` with the right arguments, mirroring what was set during build configuration.  
All parameters of `project_options()` are exposed, except `PREFIX`, since it depends on choices of the developer and not of the builder.
Using `setup_project_options()` is completely optional, and the change allows the developer to continue using project_options in classic mode.

## Possible improvements
Of course, better names could be chosen for the `setup_project_options()` macro, along with the names and descriptions of the generated options.  
Most importantly, you can consider adding parameters to `setup_project_options()` to allow the developer to decide which options to expose to the builder.

## Example
I've performed some tests forking [cpp_vcpkg_project](https://github.com/aminya/cpp_vcpkg_project) into [cpp_conan_project](https://github.com/mous16/cpp_conan_project) (sadly I can't use vcpkg right now).
The forked project uses the new `setup_project_options()` and can be configured and built with something like this:
```
> cmake `
>>         -S "${env:USERPROFILE}\Projects\cpp_conan_project\" `
>>         -B "${env:USERPROFILE}\Builds\cpp_conan_project.LLVM16\" `
>>         -G "Ninja Multi-Config" `
>>         -D CMAKE_VERBOSE_MAKEFILE:BOOL=ON `
>>         -D OPTIONS_ENABLE_CONAN:BOOL=ON `
>>         -D OPTIONS_CONAN_PROFILE:STRING="llvm16" `
>>         -D OPTIONS_ENABLE_CLANG_TIDY:BOOL=ON `
>>         -D FEATURE_TESTS:BOOL=ON ; `
>> cmake --build "${env:USERPROFILE}\Builds\cpp_conan_project.LLVM16\" -t all ; `
>> cmake --build "${env:USERPROFILE}\Builds\cpp_conan_project.LLVM16\" -t test
```


aminya (changes requested):
The request is valid. However, this is partially covered by `dynamic_project_options`. Could you instead update the code for that function? Some options are missing from it. If you update that, we won't need to add a new function.

Also see #111

mous16:
Today I learned about `dynamic_project_options`.
Now this pull requests seems really useless.

Sorry for bothering you all, I'll close this, and, if possible, will work on the resolution of #111 , that would match perfectly my use case.

PR #216, opened 2023-04-29
Added support for GCC static analysis

### LINKED ISSUE #214: Add an option to enable GCC 10+ analyzer
GCC has added an integrated static analyzer, it started it's life in GCC10 as an experimental feature, it would make sense to add an option to enable it, or just add "-fanalyzer" to the GCC warnings, it could also be treated similarly to the other analyzers like clang-tidy with the caveat that it only works when compiling with GCC 10 or newer.

See: https://gcc.gnu.org/onlinedocs/gcc/Static-Analyzer-Options.html

Project579:
With this change, it's now possible to enable the GCC 10+ static analysis, I implemented it similarly to the other static analysis tools but it adds the flags to the project_options target.

Fixes #214


aminya (approved):
Thanks for your contribution!

aminya (approved):

PR #217, opened 2023-05-02
fix: Parallel configure failure.

Project579:
When trying to configure multiple build directories in parallel project_options will try to create a symlink for compile_commands.json; multiple symlinks will be created at the same time with different targets, obviously this fails; Add a way to disable the symlink creation.
This also fixes CMake failing to configure projects with a read-only sources directory.


aminya (approved):

aminya (changes requested):
Could you make this feature optional, i.e., add an option for enabling it instead of disabling it?

Project579:
@aminya Do you want to make the "compile_commands.json" symlink disabled by default then? In the current version of the PR it stays enabled by default.

aminya:
> @aminya Do you want to make the "compile_commands.json" symlink disabled by default then? In the current version of the PR it stays enabled by default.

Yeah, it seems it is causing issues as you mentioned. I prefer that make it disabled by default to avoid any problems.

aminya (approved):

PR #224, opened 2023-05-17
feat: Support the usage of clang-tidy when cross compiling with GCC.

Project579:
Currently clang-tidy will fail to run if using GCC with a target different from the host, this modification allows us to extract the default target and sysroot from a GCC compiler in order to feed it to clang-tidy.

### DIFF
--- a/src/StaticAnalyzers.cmake
+++ b/src/StaticAnalyzers.cmake
@@ -50,6 +50,131 @@ macro(enable_cppcheck CPPCHECK_OPTIONS)
   endif()
 endmacro()
 
+function(_enable_clang_tidy_setup_cl
+  CXX_FLAGS
+  C_FLAGS
+)
+  set(CLANG_TIDY_CXX_FLAGS ${CXX_FLAGS})
+  set(CLANG_TIDY_C_FLAGS ${C_FLAGS})
+
+  if(CMAKE_CXX_STANDARD)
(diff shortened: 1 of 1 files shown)


aminya (changes requested):
It seems that the MinGW tests are failing on Windows.

Project579:
The tests seem to be passing now except for an expired certificate but that's not related to this PR.

aminya (approved):
Thanks very much! 🎉

src/StaticAnalyzers.cmake
```diff
@@ -50,6 +50,131 @@ macro(enable_cppcheck CPPCHECK_OPTIONS)
...
+      COMMAND_ERROR_IS_FATAL ANY
+    )
+    execute_process(
+      COMMAND "${CMAKE_CXX_COMPILER}" "-dumpmachine"
+      OUTPUT_VARIABLE CLANG_TIDY_CXX_FLAGS_COMPILER_TARGET
+      COMMAND_ERROR_IS_FATAL ANY
+    )
+
+    string(STRIP "${CLANG_TIDY_CXX_FLAGS_COMPILER_TARGET}" CLANG_TIDY_CXX_FLAGS_COMPILER_TARGET)
+    set(CLANG_TIDY_CXX_FLAGS_COMPILER_TARGET "--target=${CLANG_TIDY_CXX_FLAGS_COMPILER_TARGET}")
```
aminya, line 93: There is a DetectCompiler module. I am not sure if this part can be simplified using that. In any case, we can probably refactor it into that module can introduce some macros for compiler/target detection.

PR #225, opened 2023-06-10
fix: Disable `-fstack-clash-protection` on MacOS M1 when using Clang

FeignClaims:
Currently when using Clang on MacOS M1, enabling both `ENABLE_STACK_PROTECTION` and `WARNINGS_AS_ERRORS` prevents compilation:

```bash
clang-16: error: argument unused during compilation: '-fstack-clash-protection' [-Werror,-Wunused-command-line-argument]
```

That is, the `-fstack-clash-protection` part (currently) isn't supported on this environment. Since the `-fstack-protector-strong` part in `ENABLE_STACK_PROTECTION` works, I propose to fix the option instead of requiring users to disable it.

### DIFF
--- a/src/Hardening.cmake
+++ b/src/Hardening.cmake
@@ -19,7 +19,20 @@ function(
     endif()
 
     if(${ENABLE_STACK_PROTECTION})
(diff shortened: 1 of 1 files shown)


aminya (approved):
Thanks!

PR #231, opened 2023-08-06
feat: add check_sanitizers_support (#219)

### LINKED ISSUE #219: Add utility function to check if sanitizers are supported
This will simplify the main example in the readme

abeimler:
Fixes #219


src/Utilities.cmake
```diff
@@ -124,3 +124,48 @@ function(detect_architecture arch)
...
+
+# detect sanitizers support for compiler
+function(check_sanitizers_support
+        ENABLE_SANITIZER_ADDRESS
+        ENABLE_SANITIZER_UNDEFINED_BEHAVIOR
+        ENABLE_SANITIZER_LEAK
+        ENABLE_SANITIZER_THREAD
+        ENABLE_SANITIZER_MEMORY)
+  set(SANITIZERS "")
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
```
aminya, line 136: The tests are failing because the sanitizers are not supported on Windows with Clang and MinGW (out of the box).

```suggestion
if(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
```

aminya (changes requested):
Tjen

aminya (commented):
The documentation and the examples should be updated as well.

aminya (approved):
Thanks very much! 🎉

PR #233, opened 2023-08-06
install mingw with setup-cpp

abeimler:
> After adding cross-compiler to setup-cpp, update [Dockerfile.mingw](https://github.com/abeimler/project_options/blob/feature/toolchains/docker/Dockerfile.mingw#L16) so setup-cpp can install the cross-compiler (MinGW)

https://github.com/aminya/project_options/issues/180


aminya (approved):

aminya (approved):
Thank you!

PR #240, opened 2023-10-02
Add a wrapper module for add_test

FeignClaims:
from aminya/cpp_vcpkg_project#35.

Additionally, add `[WORKING_DIRECTORY <dir>]` and `[WILL_FAIL]` to specify the corresponding test property.


src/Tests.cmake
```diff
@@ -1,9 +1,380 @@
...
+  )
+  cmake_parse_arguments(args "${options}" "${one_value_args}" "${multi_value_args}" ${ARGN})
+
+  if(${type} STREQUAL "library_test")
+    add_executable(${target_name})
+    set(scope PRIVATE)
+  elseif(${type} STREQUAL "test_config")
+    add_library(${target_name} INTERFACE)
+    set(scope INTERFACE)
+  endif()
```
aminya, line 36: What happens in the else condition?

src/Tests.cmake
```diff
@@ -1,9 +1,380 @@
...
+
 # Enable coverage reporting for gcc/clang
 function(enable_coverage _project_name)
   if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
     target_compile_options(${_project_name} INTERFACE --coverage -O0 -g)
     target_link_libraries(${_project_name} INTERFACE --coverage)
   endif()
 endfunction()
+
+function(_configure_target target_name type)
```
aminya, line 13: I like how this function lumps together all the CMake functions making things more declarative. Could you name it as something like `add_target`?

src/Tests.cmake
FeignClaims: The function `_configure_target` currently only supports type `"library_test"` and `"test_config"`. Here the if-else checks the type to add the corresponding target, and set the `scope` correctly.

For instance, a `"library_test"` is an executable target, it will link all its libraries PRIVATE-ly; a `"test_config"` is an interface library as it is used to propagate the test configuration, and it will link its libraries libraries INTERFACE-ly.

src/Tests.cmake
FeignClaims: The problem here is whether I should split `DEPENDENCIES_CONFIG`, `DEPENDENCIES` and `LIBRARIES` to `[PRIVATE|PUBLIC|INTERFACE]_DEPENDENCIES_CONFIG` and so on.

Currently this function only supports `"library_test"` (executable) and `"test_config"` (header-only library), they are expected to only link one dependency type (PRIVATE for executable and INTERFACE for header-only library).

But for a non-header-only library, it might link a PRIVATE library, a PUBLIC library and a INTERFACE library at the same time. If this function is named as `add_blah`, users might expect this function also works on non-header-only libraries, so for a correct linkage, `DEPENDENCIES_CONFIG` must be split to `[PRIVATE|PUBLIC|INTERFACE]_DEPENDENCIES_CONFIG`.

FeignClaims:
Added these functions to test projects. `bext-ut` is added by vcpkg as the test framework.

FeignClaims:
Changed test projects based on this feature and tested well, ready to merge.

tests/myproj/CMakeLists.txt
```diff
@@ -97,7 +97,7 @@ add_subdirectory(libs)
 
 ## tests
 enable_testing()
-add_test(NAME main COMMAND main)
+add_executable_test(main no_arg)
```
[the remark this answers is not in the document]
FeignClaims: The `add_executable_test` actually creates a test named `test.<executable>.<test_name>`. Since users might write multiple tests for an executable, `no_arg` here is just a name saying this test has no arg.

PR #249, opened 2024-02-23
feat: support conan 2

### LINKED ISSUE #212: Support Conan 2
I am trying to enable conan and facing the following error. Also can I get conan_extra_options?

-- Conan: Adding conancenter remote repository (https://center.conan.io/) verify ssl (True)
-- Configuring incomplete, errors occurred!

## Upvote & Fund

- I am using [Polar.sh](https://polar.sh/aminya) so you can upvote and help fund this issue. The funding is received once the issue is completed & confirmed by you.

- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/aminya/project_options/issues/212">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/aminya/project_options/issues/212/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

FeignClaims:
Support conan 2 by replacing the old conan script with [`conan_provider.cmake`](https://github.com/conan-io/cmake-conan/blob/develop2/conan_provider.cmake).

Todo:

- [ ] customizing conan profiles and invocation of conan install.

Current design questions:

- The new `conan_provider.cmake` script for conan 2 requires CMake 2.24+. Is it acceptable to set `cmake_minimum_required(VERSION 3.24)` for the whole project_options?
- The cmake script for conan 2 is incompatible with conan 1. Should I keep the conan 1 function or replace it?

Close #212

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,15 +76,11 @@ jobs:
           cmake: ${{ matrix.cmake }}
           ninja: true
           vcpkg: true
-          conan: false
+          conan: 2.1.0
           cppcheck: true
           clangtidy: true
           task: true
           doxygen: ${{ !contains(matrix.os, 'macos-11') }}
-
-      - name: Install conan2
-        run: |
-          pip3 install conan==2.1.0
 
       - name: Test
         if: ${{ !cancelled() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,11 +76,15 @@ jobs:
           cmake: ${{ matrix.cmake }}
           ninja: true
           vcpkg: true
-          conan: true
+          conan: false
           cppcheck: true
           clangtidy: true
           task: true
           doxygen: ${{ !contains(matrix.os, 'macos-11') }}
+
+      - name: Install conan2
+        run: |
+          pip3 install conan==2.1.0
 
       - name: Test
         if: ${{ !cancelled() }}
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ CMake experience following the best practices.
     FetchContent, vcpkg, etc.
 -   `run_vcpkg`: automatic installation of vcpkg and the project
     dependencies
+-   `run_conan2`: automatic installation of vcpkg and the project
+    dependencies
 -   `ENABLE_CONAN` in `project_options`: automatic installation of Conan
     and the project dependencies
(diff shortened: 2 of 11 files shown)


aminya:
Can you keep the old version? Maybe we can specify the version in the Enable_conan flag (with the default being 1). 
The version can be also detected from a CLI call to Conan.

FeignClaims:
> Can you keep the old version? Maybe we can specify the version in the Enable_conan flag (with the default being 1). The version can be also detected from a CLI call to Conan.

Ok.

FeignClaims:
> So there's no way (AFAIK) to generalize them to one function.

It occurs to me there's a way to generalize them: a function `run_conan()` which should be invoked before `project(...)`. Then,

- If conan 1 is used, `run_conan()` registers a invocation. When `project_options(...)` is invoked, it checks the registration and actually calls `_run_conan1()` if it should.
- If conan 2 is used, `run_conan()` invokes `_run_conan2()`.

This is somehow wired since conan 1 actually runs when `project_options(...)` is invoked.

Should I switch to use this generalized function or just separate them into different functions like now?

.github/workflows/ci.yml
```diff
@@ -76,12 +76,16 @@ jobs:
           cmake: ${{ matrix.cmake }}
           ninja: true
           vcpkg: true
-          conan: true
+          conan: false
```
aminya, line 79: You can instead specify the version here. No need for pip

aminya (changes requested):

aminya (commented):
I like the idea of run_conan. Should we deprecate the Enable_conan option?

.github/workflows/ci.yml
FeignClaims: done

FeignClaims:
Finished `run_conan()` and ready to merge.

> Should we deprecate the Enable_conan option?

I remove it in documentation but keep t it in code for backward compability. A `DEPRECATED_CALL` option is added to distinguish the version of the call, That is,

- `run_conan()` before `project(...)`: supports both conan 1 and conan 2.
- `ENABLE_CONAN` in `project_options(...)` after `project(...)`: removed from documentation, only supports conan 1, messages a FATAL_ERROR if users enable it using conan 2.

aminya (approved):
Thanks a lot for your contribution! 💯

PR #252, opened 2024-03-11
Update WindowsToolchain to v0.9.0

ClausKlein:


src/CrossCompiler.cmake
```diff
@@ -356,7 +356,7 @@ macro(enable_cross_compiler)
     else()
       if(NOT DEFINED EMSCRIPTEN_ROOT)
         include(FetchContent)
-        message(STATUS "fetch emscripten repo. ...")
+        message(WARNING "fetch emscripten repo main branch. ...")
```
aminya, line 359: I am not sure why this is a warning
```suggestion
        message(STATUS "fetch emscripten repo main branch. ...")
```

aminya (changes requested):

src/CrossCompiler.cmake
ClausKlein: I want to show that is a moving target!

PR #255, opened 2024-03-27
fix: Disable gcc sanitizer on macos by default

FeignClaims:
According to [this comment](https://github.com/orgs/Homebrew/discussions/3384#discussioncomment-6264292), gcc now dosen't support libsanitizer on macOS with arm64 and macOS Ventuna or later with x86_64.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,10 @@ jobs:
             compiler: "gcc"
             cmake: true
             vcvarsall: true
+          - os: "macos-13"
+            compiler: "gcc"
+            cmake: true
+            vcvarsall: true
         exclude:
           # fails with an internal error
           - os: "macos-12"
@@ -80,7 +84,7 @@ jobs:
           cppcheck: true
           clangtidy: true
           task: true
-          doxygen: ${{ !contains(matrix.os, 'macos-11') }}
+          doxygen: ${{ !contains(matrix.os, 'macos-11') && !contains(matrix.os, 'macos-13') }}
 
       - name: Test
         if: ${{ !cancelled() }}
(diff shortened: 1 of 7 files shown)


aminya (changes requested):
But this issue didn't happen inside CI. We should be able to reproduce it to disable it?

FeignClaims:
> But this issue didn't happen inside CI. We should be able to reproduce it to disable it?

The issue happens inside the DynamicProjectOptions.cmake, which sets UBSAN and ASAN for gcc by default (except for windows):

> ```cmake
>   if((CMAKE_CXX_COMPILER_ID MATCHES ".*Clang.*" OR CMAKE_CXX_COMPILER_ID MATCHES ".*GNU.*")
>      AND WIN32
>   )
>    ...
>  else()
>    ...
>  endif()
> ```

So use `dynamic_project_options` instead of `project_options` should trigger the issue, but I failed to figure out how to add such a  test.

FeignClaims:
Convert to a draft until

- find a way to test this.
- detect MacOS OS version to keep gcc sanitizers enable on x86_64 macOS Monterey or older.

FeignClaims:
@aminya The issue happened in the new gcc macos-13 [ci](https://github.com/aminya/project_options/actions/runs/8702993820/job/23868245200) as expected.

FeignClaims:
Ready to merge.

changes:

- Disable gcc sanitizers on x86_64 MacOS 13+ and arm64 MacOS.
- Add tests for this by adding a gcc macos-13 ci run and enabling sanitizers for it. (Sadly, I can only test arm64 MacOS munally as github actions dosen't support it.)
- For consistency, use `detect_sanitizers_support` inside `dynamic_project_options` to detect whether enabling sanitizers or not.

aminya (approved):

aminya:
Thanks for the investigation. It would be great if you could create an issue for the upstream so that they can look into the problem.

FeignClaims:
Although no related issue, it seems that the upstream already knows the issue and chooses not to support this for now (see https://github.com/iains/gcc-darwin-arm64/commit/e722a1f42b28092c9f709a3f758fc4fe57db32b0).

PR #264, opened 2024-06-17
fix(MSVC): Don't enforce /Zi if POLICY CMP0141 is available

### LINKED ISSUE #262: Enabling Address Sanitizer overrides the debug format on MSVC
I'm trying to build a static library with embedded debug information on MSVC with Address Sanitizer enabled, and I'm seeing the following warning: `cl : Command line warning D9025 : overriding '/Z7' with '/Zi'`
The reason this happens is this line, it sets the `/Zi` flag unconditionally: https://github.com/aminya/project_options/blob/46081b21083cccf6a3be6bc8b73d3f6b39eb4964/src/Sanitizers.cmake#L90

In my code I'm setting the debug information format like this:
 ```cmake
set_target_properties(<target> PROPERTIES MSVC_DEBUG_INFORMATION_FORMAT Embedded)
```

## Upvote & Fund

@aminya is using [Polar.sh](https://polar.sh/aminya) so you can upvote and help fund this issue. The funding is received once the issue is completed & confirmed by you.

Thank you in advance for helping prioritize & fund our backlog!

<br/>

<a href="https://polar.sh/aminya/project_options/issues/262">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/aminya/project_options/issues/262/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

### PR DESCRIPTION
**Pull request by ddassie-texa**
With the current implementation if CMAKE_MSVC_DEBUG_INFORMATION_FORMAT is set, it won't be respected as the /Zi flag set on enable_sanitizers will always override it, just skip setting it if POLICY CMP0141 is available, if the user has set CMAKE_MSVC_DEBUG_INFORMATION_FORMAT to EditAndContinue, than only in that case override it.

Fixes #262

### DIFF
--- a/src/Sanitizers.cmake
+++ b/src/Sanitizers.cmake
@@ -86,8 +86,19 @@ function(
               "Using MSVC sanitizers requires setting the MSVC environment before building the project. Please manually open the MSVC command prompt and rebuild the project."
           )
         endif()
+        if(POLICY CMP0141)
+          if("${CMAKE_MSVC_DEBUG_INFORMATION_FORMAT}" STREQUAL ""
+             OR "${CMAKE_MSVC_DEBUG_INFORMATION_FORMAT}" STREQUAL "EditAndContinue"
+           )
+            set_target_properties(
+              ${_project_name} PROPERTIES MSVC_DEBUG_INFORMATION_FORMAT ProgramDatabase
+            )
+          endif()
+        else()
+          target_compile_options(${_project_name} INTERFACE /Zi)
+        endif()
         target_compile_options(
-          ${_project_name} INTERFACE /fsanitize=${LIST_OF_SANITIZERS} /Zi /INCREMENTAL:NO
+          ${_project_name} INTERFACE /fsanitize=${LIST_OF_SANITIZERS} /INCREMENTAL:NO
         )
         target_link_options(${_project_name} INTERFACE /INCREMENTAL:NO)
       endif()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Sanitizers.cmake [hunk 1]
```diff
@@ -86,8 +86,17 @@ function(
...
+            set_target_properties(
+              ${_project_name} PROPERTIES MSVC_DEBUG_INFORMATION_FORMAT ProgramDatabase
+            )
+          endif()
+        else()
+          target_compile_options(${_project_name} INTERFACE /Zi)
+        endif()
         target_compile_options(
-          ${_project_name} INTERFACE /fsanitize=${LIST_OF_SANITIZERS} /Zi /INCREMENTAL:NO
+          ${_project_name} INTERFACE /fsanitize=${LIST_OF_SANITIZERS} /INCREMENTAL:NO
```
aminya, line 99: Would this break the default sanitizer experience? We want the sanitizer to work by default in MSVC.

ddassie-texa: I don't think so, from version 3.25 CMake sets `MSVC_DEBUG_INFORMATION_FORMAT` to `$<$<CONFIG:Debug,RelWithDebInfo>:ProgramDatabase>`, (see: https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_DEBUG_INFORMATION_FORMAT.html) it might break for `Release/MinSizeRel` builds as no debug info is generated by default.
We could set `MSVC_DEBUG_INFORMATION_FORMAT` to `ProgramDatabase` unconditionally, users can still override it in their own targets, but if they have explicitly set `CMAKE_MSVC_DEBUG_INFORMATION_FORMAT` to a compatible value we should not override it.

aminya: I just want to make sure we don't have a regression in the default experience. So as long as that's the case, this change should be OK.

PR #271, opened 2024-08-13
feat: Update default language standard to the latest standard + feat: Set cppcheck default std instead of defaulting standard based on it

FeignClaims:
Update the default `CMAKE_CXX_STANDARD` and `CMAKE_C_STANDARD` settings to the latest standard version if they're not specified by the user.

Additionally, the standard test code is generalized to a function `_set_language_standard`:

```cmake
_set_language_standard(CXX_LATEST_STANDARD CXX 23 20 17 14 11)
_set_language_standard(C_LATEST_STANDARD C 23 20 17 11 99 90)
```


FeignClaims:
C++ fails to use 23 and C fails to use 23,20,17 because cppcheck dosen't support them. Instead of implicitly using an old standard, I think it's fairly reasonable to just use the latest standard while setting cppcheck to use the default standard, and give a `${WARNING_MESSAGE}` about the cppcheck behaviour to those who care.

Also, the corresponding cppcheck option of `set(CMAKE_C_STANDARD 90)` is `--std=c89` according to `cppcheck --help`:

```txt
--std=<id>           Set standard.
                     The available options are:
                      * c89
                             C code is C89 compatible
                      * c99
                             C code is C99 compatible
                      * c11
                             C code is C11 compatible (default)
                      * c++03
                             C++ code is C++03 compatible
                      * c++11
                             C++ code is C++11 compatible
                      * c++14
                             C++ code is C++14 compatible
                      * c++17
                             C++ code is C++17 compatible
                      * c++20
                             C++ code is C++20 compatible (default)
```

Versions are hard-coded in the code, as cppcheck runs normally even if I run `cppcheck --std=c++100` so I can't test whether the option is correct by running it.

```cmake
if(CMAKE_CXX_STANDARD IN_LIST 03 11 14 17 20)
  set(CMAKE_CXX_CPPCHECK ${CMAKE_CXX_CPPCHECK} --std=c++${CMAKE_CXX_STANDARD})
else()
  message(${WARNING_MESSAGE} "cppcheck dosen't support specified C++ standard ${CMAKE_CXX_STANDARD}. Using the cppcheck default C++ standard version.")
endif()
```

src/StaticAnalyzers.cmake
```diff
@@ -38,11 +38,21 @@ macro(enable_cppcheck CPPCHECK_OPTIONS)
     set(CMAKE_C_CPPCHECK ${CMAKE_CXX_CPPCHECK})
 
     if(NOT "${CMAKE_CXX_STANDARD}" STREQUAL "")
-      set(CMAKE_CXX_CPPCHECK ${CMAKE_CXX_CPPCHECK} --std=c++${CMAKE_CXX_STANDARD})
+      if(CMAKE_CXX_STANDARD MATCHES [[03|11|14|17|20]])
+        set(CMAKE_CXX_CPPCHECK ${CMAKE_CXX_CPPCHECK} --std=c++${CMAKE_CXX_STANDARD})
+      else()
```
aminya, line 43: What if cppcheck adds support for C++23?

src/StaticAnalyzers.cmake
```diff
@@ -38,11 +38,21 @@ macro(enable_cppcheck CPPCHECK_OPTIONS)
...
+        message(${WARNING_MESSAGE} "cppcheck dosen't support specified C++ standard ${CMAKE_CXX_STANDARD}. Using the cppcheck default C++ standard version.")
+      endif()
     endif()
 
     if(NOT "${CMAKE_C_STANDARD}" STREQUAL "")
-      set(CMAKE_C_CPPCHECK ${CMAKE_C_CPPCHECK} --std=c${CMAKE_C_STANDARD})
+      if(CMAKE_C_STANDARD MATCHES [[90]])
+        set(CMAKE_C_CPPCHECK ${CMAKE_C_CPPCHECK} --std=c89)
+      elseif(CMAKE_C_STANDARD MATCHES [[99|11]])
```
aminya, line 51: Same here about C17

src/StaticAnalyzers.cmake
```diff
@@ -54,14 +64,6 @@ function(_enable_clang_tidy_setup_cl CXX_FLAGS C_FLAGS)
   set(CLANG_TIDY_CXX_FLAGS ${CXX_FLAGS})
   set(CLANG_TIDY_C_FLAGS ${C_FLAGS})
 
-  if(CMAKE_CXX_STANDARD)
```
aminya, line 57: Why clang tidy args are removed?

aminya (changes requested):

src/StaticAnalyzers.cmake
aminya: > Versions are hard-coded in the code, as cppcheck runs normally even if I run cppcheck --std=c++100 so I can't test whether the option is correct by running it.

That's a good reason.

src/StaticAnalyzers.cmake
FeignClaims: I have to remove it, otherwise clang-tidy wrongly sets`-std=c++23` according to `CMAKE_CXX_STANDARD` while clang 15 only supports `-std=c++2b` ([failure example](https://github.com/aminya/project_options/actions/runs/10371430640/job/28711716651#step:5:1380)). This change is ok, since clang-tidy knows the actual compile flags (including `-std=<version>`) according to [the cmake implementation code](https://gitlab.kitware.com/cmake/cmake/-/blob/8733dcd11588f2d9eb0bddc8af8a6db7a2e9eba8/Source/cmcmd.cxx#L370-416), especially [this comment](https://gitlab.kitware.com/cmake/cmake/-/blob/8733dcd11588f2d9eb0bddc8af8a6db7a2e9eba8/Source/cmcmd.cxx#L382-389).

src/StaticAnalyzers.cmake
FeignClaims: That makes sense. I'll give it another try in a few days.

aminya:
Is this ready to get merged? Can we address the essential part of my comments, or do you think this will not cause issues?

src/StaticAnalyzers.cmake
aminya: I think the current implementation is good enough for now. Detecting the supported standard is hard on our side. Cppcheck should probably provide some flags

FeignClaims: I think so too. Perhaps I can manually maintain version issues afterwards.

FeignClaims:
> Is this ready to get merged? Can we address the essential part of my comments, or do you think this will not cause issues?

Sorry for the delay. According to my recent attempts, cppcheck itself is unable to detect the version issue, so this is the best result we can achieve at the moment.

I believe the current changes are sufficient and ready to merge, and I can manually maintain the version changes afterwards - after all, we were manually adjusting to the latest C++ version before.

aminya (approved):
Thank you for the contribution

PR #279, opened 2024-09-11
feat: install runtimes for packages installed by vcpkg

### LINKED ISSUE #261: feature request: support installing imported targets for vcpkg and conan
Although not universally appicable to all `IMPORTED` targets as #167 is trying to do, vcpkg and conan 2.4.0 already provide a way to install their `IMPORTED` targets:

- vcpkg provides two options for users to easily copy ~~runtimes~~ (only) dlls to the `build directory` and `installation directory`. (learnt from [this comment](https://github.com/conan-io/conan/issues/12654#issuecomment-1946135006))

- conan 2.4.0's CMakeToolchain generator provides a cmake variable `CONAN_RUNTIME_LIB_DIRS` that allows users to install runtime libraries. (see the corresponding documentation [here](https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#conan-runtime-lib-dirs), and [this comment](https://github.com/conan-io/conan/issues/12654#issuecomment-2326131267) shows how to copy runtimes to output folder with it)

So maybe we could support them respectively? (or unify them into a general function?)

- [ ] support installing runtimes for vcpkg #279 
- [ ] support installing runtimes for conan #281

FeignClaims:
~~resolves #261~~

resolves the vcpkg part of #261

### DIFF
--- a/.github/workflows/ci.cross.arm.yml
+++ b/.github/workflows/ci.cross.arm.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
   Test:
-    if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.cross.mingw.yml
+++ b/.github/workflows/ci.cross.mingw.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
   Test:
-    if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
--- a/.github/workflows/ci.emscripten.yml
+++ b/.github/workflows/ci.emscripten.yml
@@ -8,7 +8,6 @@ on:
 
 jobs:
   Test:
-    if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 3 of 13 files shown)


FeignClaims:
Although this works well on my MacBook, it clearly needs to be tested with CI. However, I'm not sure how to write tests for it. Could you provide some suggestions? @aminya

aminya:
> ~~Although this works well on my MacBook, it clearly needs to be tested with CI. However, I'm not sure how to write tests for it. Could you provide some suggestions? @aminya ~~
> 
> I choose to simply run the installed executable after installation.

The tests should be good. We can also test it in a Docker container to make sure it runs in a clean system.

FeignClaims:
The vcpkg part is done.

Blocked by conan because it only supports runtime installation when it is invoked separately, as opposed to including conan_provider.cmake in CMakeLists.txt. conan-io/conan#17018

FeignClaims:
> Blocked by conan because it only supports runtime installation when it is invoked separately, as opposed to including conan_provider.cmake in CMakeLists.txt. https://github.com/conan-io/conan/issues/17018

According to the maintainer, conan might add the support in 2.9 (currently 2.7.1). Let's merge the vcpkg part first and merge the conan part once it gets supported.

FeignClaims:
> We can also test it in a Docker container to make sure it runs in a clean system.
https://github.com/aminya/setup-cpp/blob/691fa792a8d5fd90d3669024745a5ab6572c80ad/dev/docker/examples/ubuntu-llvm.dockerfile#L9

@aminya I'm not familiar enough with Docker and the test framework of this project, so I'm not able to write such a docker test. Could you finish this part of test? The rest is ready to merge btw.

aminya (approved):

PR #283, opened 2024-10-04
fix: enable sanitizer correctly

FeignClaims:
Since #278, sanitizers are not enabled because we check the wrong letter case (for instance, `ENABLE_SANITIZER_address` instead of `ENABLE_SANITIZER_ADDRESS`). This pr fixes that.

In addition, we check `ENABLE_SANITIZER_undefined`, but the interface is `ENABLE_SANITIZER_UNDEFINED_BEHAVIOR`. For this option, let's replace `ENABLE_SANITIZER_UNDEFINED_BEHAVIOR` with `ENABLE_SANITIZER_UNDEFINED`, and warn that `UNDEFINED_BEHAVIOR` is deprecated.


aminya (approved):
Oops. Thanks for catching this

